PR #6, opened 2015-08-25
radosgw: Deploy radosgw using civetweb

tserong:
This is approximately what would need to be done to deploy radosgw using
civetweb, instead of apache.

Outstanding Issues:
- SSL doesn't presently work due to unversioned libssl.so and
  libcrypto.so being dlopen()'d (see http://tracker.ceph.com/issues/11239
  and https://bugzilla.suse.com/show_bug.cgi?id=942874), but you can
  fake it up with a couple of symlinks to test.
- Discrepancy between apache-style separate SSL key and certfiles, and
  civetweb combined pemfile.  I'm not sure how best to address that (see
  question in source).

Also, the radosgw_apache2.rb recipe and related templates should be
deleted.


tboerger:
LGTM

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -31,6 +31,25 @@
...
+    rgw_port = rgw_use_ssl ? node["ceph"]["ha"]["ports"]["radosgw_ssl"] : node["ceph"]["ha"]["ports"]["radosgw_plain"]
+  else
+    rgw_port = rgw_use_ssl ? node["ceph"]["radosgw"]["rgw_port_ssl"] : node["ceph"]["radosgw"]["rgw_port"]
+  end
+  if rgw_use_ssl
+    # This is a combined cert + key, unlike with apache which have separate
+    # files.  This would need to be exposed in the barlcamp UI, much as the
+    # existing certificate and key are.
+    # QUESTION: Should we keep the existing cert & key attributes, and
+    # concatenate them automatically to make the pemfile?
```
vuntz: Yes, I would concatenate them automatically.

vuntz:
@tserong do you feel we should create the symlinks for the libs in the cookbook?

tserong:
> @tserong do you feel we should create the symlinks for the libs in the cookbook?

No, those symlinks are a bad idea. This needs to be fixed in the radosgw build so it uses proper dynamic linking - see https://bugzilla.suse.com/show_bug.cgi?id=942874#c7

vuntz:
So the issue with the libs is fixed with https://github.com/SUSE/ceph/pull/18.

tserong:
> So the issue with the libs is fixed with SUSE/ceph#18.

Thanks @vuntz, I'll get back onto this.

tserong:
OK, this should be about right now. The one thing I'm not set up to test ATM is an HA deployment, and I note that line 49 of `chef/cookbooks/ceph/recipes/radosgw_ha.rb` says `include_recipe "crowbar-pacemaker::apache"`. Is this of any significance?

tserong:
Jeez, Hound is a picky bugger, isn't he?

chef/cookbooks/ceph/recipes/radosgw_civetweb.rb
```diff
@@ -0,0 +1,78 @@
...
+  # may not actually exist, hence the existence check below
+  ruby_block "Creating radosgw pemfile" do
+    block do
+      `cp #{certfile} #{pemfile}`
+      `cat #{keyfile} >> #{pemfile}` if ::File.exist? keyfile
+      FileUtils.chown "root", node["ceph"]["radosgw"]["group"], pemfile
+      FileUtils.chmod 0640, pemfile
+    end # block
+  end # ruby_block
+end
```
tserong: I think Hound is lying about this.

tserong:
So, @vuntz, @tboerger, how's this look now?

chef/cookbooks/ceph/recipes/radosgw_civetweb.rb
mjura: Woof, woof :)

chef/cookbooks/ceph/recipes/radosgw_civetweb.rb
```diff
@@ -0,0 +1,78 @@
+# This recipe doesn't do anything other than deal with SSL keys and certificates.
+# Everything else for civetweb setup is in conf.rb.  It could arguably be merged
+# into radosgw.rb; it only exists as a separate file because it used to be
+# radosgw_apache2.rb
+
+if node["ceph"]["radosgw"]["ssl"]["enabled"]
```
mjura: Try, maybe this help for Hound

```
return unless node["ceph"]["radosgw"]["ssl"]["enabled"]
```

chef/cookbooks/ceph/recipes/radosgw_civetweb.rb
mjura: Look comment above

vuntz:
Looks good to me, let's see what gating says.

chef/cookbooks/ceph/recipes/radosgw_civetweb.rb
tserong: If nobody objects, I'd rather just leave it as is for the moment and live with the Hound warning, to save me having to test it again ;-)

tserong:
How long does gating take? (Not that I'm in an enormous hurry or anything, but it's been yellow for 24 hours now, I think)

chef/cookbooks/ceph/recipes/radosgw_civetweb.rb
mjura: I'm sad then, it would simplify this recipe and decrease indentation.

crowbar_framework/app/views/barclamp/ceph/_edit_attributes.html.haml
```diff
@@ -12,7 +12,7 @@
       %legend
         = t(".radosgw.ssl.header")
 
-      = boolean_field %w(radosgw ssl enabled), :collection => :radosgw_ssl_protocols, "data-sslprefix" => "radosgw_ssl", "data-sslcert" => "/etc/apache2/ssl.crt/ceph-radosgw.crt", "data-sslkey" => "/etc/apache2/ssl.key/ceph-radosgw.key"
+      = boolean_field %w(radosgw ssl enabled), :collection => :radosgw_ssl_protocols, "data-sslprefix" => "radosgw_ssl", "data-sslcert" => "/etc/ceph/ceph-radosgw.crt", "data-sslkey" => "/etc/ceph/ceph-radosgw.key"
```
mjura: Can we break this line ?

vuntz:
> How long does gating take?

It was apparently stuck; I re-triggered it.

tserong:
Last commit special for @mjura ;-)

mjura:
I was testing this change and works like charm, finally don't have to worry about Apache2

+1 from me

jsuchome:
Originally, after deploying radosgw role, swift CLI was working out of the box, we were testing it with this code: https://github.com/SUSE-Cloud/automation/blob/master/scripts/qa_crowbarsetup.sh#L2983

Does it work now?

jsuchome:
Oh, I've just seen it working on @mjura machine, so yes, it does work (with --insecure option; I have no idea how to fix SSL problems btw).

+1

PR #7, opened 2015-09-10
Merging existing client caps into new one, (bsc#944810)

mjura:
This is fixing https://bugzilla.suse.com/show_bug.cgi?id=944810

### DIFF
--- a/chef/cookbooks/ceph/providers/client.rb
+++ b/chef/cookbooks/ceph/providers/client.rb
@@ -8,7 +8,7 @@ action :add do
   admin_keyring = @new_resource.admin_keyring
   filename = @current_resource.filename
   keyname = @current_resource.keyname
-  caps = @new_resource.caps.map { |k, v| "#{k} '#{v}'" }.join(" ")
+  caps = @new_resource.caps
   owner = @new_resource.owner
   group = @new_resource.group
   mode = @new_resource.mode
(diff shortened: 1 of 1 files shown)


chef/cookbooks/ceph/providers/client.rb
```diff
@@ -60,13 +60,20 @@ def get_caps(ceph_conf, admin_keyring, keyname)
 
 def auth_set_key(ceph_conf, admin_keyring, keyname, caps)
   # try to add the key
-  cmd = "ceph -k #{admin_keyring} -c #{ceph_conf} auth get-or-create #{keyname} #{caps}"
+  caps_tmp = {}
```
vuntz: Not useless, but I think there's a point that the variable name is not great.

chef/cookbooks/ceph/providers/client.rb
```diff
@@ -60,13 +60,21 @@ def get_caps(ceph_conf, admin_keyring, keyname)
...
   get_or_create = Mixlib::ShellOut.new(cmd)
   get_or_create.run_command
   if get_or_create.stderr.scan(/EINVAL.*but cap.*does not match/)
     Chef::Log.info("Deleting old key with incorrect caps")
     # delete an old key if it exists and is wrong
     Mixlib::ShellOut.new("ceph -k #{admin_keyring} -c #{ceph_conf} auth del #{keyname}").run_command
+    # don't remove custom existing caps for osd pools
+    caps_osd = ""
+    caps_osd  = caps["osd"] + "," + get_caps(ceph_conf, admin_keyring, keyname)["osd"]
+    caps["osd"] = caps_osd.split(",").collect { |x| x.strip }.uniq.join(",")
```
tboerger:
``` ruby
caps_osd.split(",").collect(&:strip).uniq.join(",")
```

tboerger:
Beside the single comment it looks ok codewise.

chef/cookbooks/ceph/providers/client.rb
mjura: Thanks @tboerger, I was searching how to pass this `strip` as argument to block.

chef/cookbooks/ceph/providers/client.rb
mjura: @vuntz : How about `caps_str`, which is more suitable.

mjura:
Ok, change is ready. Was tested with different configuration and works fine. It will never overwrite custom caps for external ceph integration. Instead of this it will extend caps for osd with new one.

rsalevsky:
+1 but please wait for the gating

chef/cookbooks/ceph/providers/client.rb
```diff
@@ -60,11 +60,17 @@ def get_caps(ceph_conf, admin_keyring, keyname)
...
-  cmd = "ceph -k #{admin_keyring} -c #{ceph_conf} auth get-or-create #{keyname} #{caps}"
+  caps_str = caps.map { |k, v| "#{k} '#{v}'" }.join(" ")
+  cmd = "ceph -k #{admin_keyring} -c #{ceph_conf} auth get-or-create #{keyname} #{caps_str}"
   get_or_create = Mixlib::ShellOut.new(cmd)
   get_or_create.run_command
-  if get_or_create.stderr.scan(/EINVAL.*but cap.*does not match/)
+  unless get_or_create.stderr.scan(/EINVAL.*but cap.*does not match/).empty?
     Chef::Log.info("Deleting old key with incorrect caps")
+    # don't remove custom existing caps for osd pools
+    caps_osd = caps["osd"] + "," + get_caps(ceph_conf, admin_keyring, keyname)["osd"]
```
vuntz: Asking out loud: does the order matter? (I don't think it does, just checking). If it does, then `get_caps` needs to go first.

mjura: It doesn't matter, because afterwards it is splitted and sorted. But I agree with comment and I will change this order

tboerger:
LGTM

jsuchome:
+1 (based on code)

PR #12, opened 2015-09-30
replace check for platform with check for platform_family

jsuchome:


chef/cookbooks/ceph/attributes/radosgw.rb
```diff
@@ -30,14 +30,14 @@
...
-  else
-    default["ceph"]["radosgw"]["service_name"]  = "ceph-radosgw"
+when "ubuntu"
+  default["ceph"]["radosgw"]["service_name"] = "radosgw-all-starter"
+when "debian"
+  default["ceph"]["radosgw"]["service_name"] = "radosgw"
+when "suse", "opensuse"
+  default["ceph"]["radosgw"]["service_name"] = "ceph-radosgw@#{node["hostname"]}"
+else
+  default["ceph"]["radosgw"]["service_name"] = "ceph-radosgw"
```
vuntz, line 41: Somehow, I feel we should move more towards `platform_family` (to avoid bad surprises for new distros). What about something like:

``` ruby
case node["platform_family"]
  when "debian"
    if node["platform"] == "ubuntu"
      default["ceph"]["radosgw"]["service_name"]  = "radosgw-all-starter"
    else
      default["ceph"]["radosgw"]["service_name"]  = "radosgw"
    end
  when "suse"
    default["ceph"]["radosgw"]["service_name"]  = "ceph-radosgw@#{node["hostname"]}"
  else
    default["ceph"]["radosgw"]["service_name"]  = "ceph-radosgw"
end
```

vuntz:
Suggestions in https://github.com/jsuchome/crowbar-ceph/pull/1 -- feel free to squash that in your commit.

vuntz:
+1

mjura:
Looks ok +1

aplanas:
+1

PR #13, opened 2015-10-02
Unbreak Ceph again

dirkmueller:
We do want to set a mon attribute not an osd attribute.


vuntz:
+1

jsuchome:
+1

tboerger:
:+1:

mjura:
+1 
Should we wait for gating tests?

dirkmueller:
@mjura since the previous change(s) were merged without testing, I'd wait to see that I found all regressions introduced in the last change.

PR #14, opened 2015-10-02
Suppress warning  mon_pg_warn_max_per_osd, (bnc#948375)

mjura:
In Ceph Hammer release was introduced new warning
  health HEALTH_WARN
         too many PGs per OSD (.... > max 300)

For small RADOS cluster we would like to disable it
This is also tracked by https://bugzilla.suse.com/show_bug.cgi?id=948375

### DIFF
--- a/chef/cookbooks/ceph/templates/default/ceph.conf.erb
+++ b/chef/cookbooks/ceph/templates/default/ceph.conf.erb
@@ -33,6 +33,14 @@
         ; default number of placement groups for placement for a pool
         osd pool default pgp num = <%= @pool_pg_num %>
 
+        # New warning was introduced in Ceph Hammer release
+        # For small RADOS clusters with less than 5 OSDs 
+        # and pools for glance, radosgw, we are getting 
+        # health HEALTH_WARN
+        #        too many PGs per OSD (1216 > max 300)
+        # https://bugzilla.suse.com/show_bug.cgi?id=948375
+        mon_pg_warn_max_per_osd = 0
+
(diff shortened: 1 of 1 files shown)


dirkmueller:
LGTM, assuming it helps :)

vuntz:
It does help, but it's really only hiding the issue under the carpet...

+1 for now, but we should get a real fix at some point.

tserong:
I agree with @vuntz, +1 for now, but something better in future (see bnc#948522)

osynge:
Thing is we have no perfect value depending on number of OSD's.

lmb suggested this link http://ceph.com/pgcalc/ 

For ceph-deploy I took a less optimal solution so that 3 OSD clusters would not show the warning of 16 PG per pool.

LMB got the above tool to work (did not allow me to adjust number of OSD's) saying

<quote>
For a radosgw only deployment, it suggests 512 PGs for the rgw bucket pool, and 4 for all others. Creating all of them with 128 PGs for sure is on the high side for the others and quite low for the bucket pool.
</quote>

Please note setting the number too low is better than setting the number too high.

mjura:
> For a radosgw only deployment, it suggests 512 PGs for the rgw bucket pool, and 4 for all others. Creating 
> all of them with 128 PGs for sure is on the high side for the others and quite low for the bucket pool.

@osynge What would be recommendation for pg_num for each pool for radosgw
.rgw.root, .rgw.control, .rgw, .rgw.gc, .users.uid, .users, .rgw.buckets.index, .rgw.buckets

PR #15, opened 2015-10-05
Allow Ceph only on suse-12.0 nodes

dirkmueller:
SES2 can not be deployed on SLE12 SP1, so require SLE12 SP0


jsuchome:
+1

aplanas:
+1

Do we need to change qa_crowbarsetup.sh to follows these restrictions?

toabctl:
+1

mjura:
+1

PR #16, opened 2015-10-09
Remove ceph-cinder and ceph-glance roles

vuntz:
There's no openstack-specific code here anymore.


mjura:
Wow, I've forgotten that we had something like this!
+1

tboerger:
:+1:

PR #20, opened 2015-11-05
Authentication to allow access to radosgw admin

mmnelemane:
Updating caps for admin tenant ID to access radosgw objects from ceilometer  through radosgw-admin.
Config file updated to accept  admin role.

### DIFF
--- a/chef/cookbooks/ceph/templates/default/ceph.conf.erb
+++ b/chef/cookbooks/ceph/templates/default/ceph.conf.erb
@@ -148,8 +148,11 @@
   log file = /var/log/radosgw/radosgw.log
   rgw frontends = civetweb port=<%= @rgw_port %><% if (@rgw_pemfile) %>s ssl_certificate=<%= @rgw_pemfile %><% end %>
   <% unless node[:ceph][:keystone_instance].nil? || node[:ceph][:keystone_instance].empty? -%>
+  rgw s3 auth use keystone = true 
   rgw keystone url =  <%= @keystone_settings['admin_auth_url'] %>
   rgw keystone admin token = <%= @keystone_settings['admin_token'] %>
+  ### default rgw keystone accepted roles : admin, Member, _member_
+  rgw keystone accepted roles = "rgw_admin, Member, _member_ "
   nss db path = <%= node['ceph']['radosgw']['nss_directory'] %>
   <% end -%>
 <% if (! node['ceph']['config']['rgw'].nil?) -%>


chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -61,6 +64,20 @@
   include_recipe "ceph::radosgw_keystone"
 end
 
+admin_tenant_id = keystone_settings["admin_tenant_id"]
+
+# add caps to user admin to access the bucketst 
+if node.roles.include?("ceph-radosgw")
```
rhafer: I think this is reduntant. As far as I can see the radosgw recipe is only ever run on nodes that actually have the radosgw-role.

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -61,6 +64,20 @@
...
+  cmd = "radosgw-admin caps add --uid=#{admin_tenant_id} --caps='buckets=*'"
+  set_bucket_caps = Chef::ShellOut.new(cmd)
+  Chef::Log.debug("Executing command: #{cmd}")
+  set_bucket_caps.run_command
+  cmd = "radosgw-admin caps add --uid=#{admin_tenant_id} --caps='users=*'"
+  set_users_caps = Chef::ShellOut.new(cmd)
+  Chef::Log.debug("Executing command: #{cmd}")
+  set_users_caps.run_command
```
rhafer: You should be using `execute` resources here. Instead of using Chef::ShellOut. Something like:

```
execute "Set radosgw bucket caps" do
    command "radosgw-admin caps add --uid=#{admin_tenant_id} --caps='buckets=*'
end
```

Probably with a proper `not_if` or `only_if` conditions to check if those caps already exist.

rhafer: Also I am pretty sure that `read` caps (instead of `*`) should be enough here. (But I didn't verify that)

rhafer: Do you happen to know where radosgw stores that caps information? Is it distributed among all radosgw nodes (e.g. incase of an HA setup) or is that a per node thing? Depending on that you might want to make sure that this code is only executed on one node.

mmnelemane: Thanks @rhafer, I have not considered HA deployment yet.  I want to make those changes in a separate PR (if its ok) after testing on a HA setup. I will make the changes for execute calls and update here.

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -35,6 +35,8 @@
 keystone_settings = {}
 if is_rgw && !(node[:ceph][:keystone_instance].nil? || node[:ceph][:keystone_instance].empty?)
   keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+  keystone_settings["s3_auth"] = true
+  keystone_settings["accepted_roles"] = ["admin", "Member", "_member_"]
```
mjura: I think that better would be to use ceilometer role instead of admin

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -45,6 +45,8 @@
 keystone_settings = {}
 if is_rgw && !(node[:ceph][:keystone_instance].nil? || node[:ceph][:keystone_instance].empty?)
   keystone_settings = KeystoneHelper.keystone_settings(node, @cookbook_name)
+  keystone_settings["s3_auth"] = true
+  keystone_settings["accepted_roles"] = ["rgw_admin", "Member", "_member_"]
```
dirkmueller: Hmm, this looks slightly awkward. (modifying the return value of a shared keystone settings helper). IMHO since this is only used in the config file template, I'd just set those values explicitely in the config template and not bother putting that into the keystone settings hash.

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -61,6 +61,27 @@
...
+
+    caps_bucket = Mixlib::ShellOut.new("radosgw-admin caps add --uid=#{uid} --caps='buckets=*'")
+    caps_bucket.run_command
+    caps_bucket.error!
+
+    caps_users = Mixlib::ShellOut.new("radosgw-admin caps add --uid=#{uid} --caps='users=*'")
+    caps_users.run_command
+    caps_users.error!
+  end
+  action :create
```
dirkmueller: isn't there some only_if / not_if missing? otherwise this causes an error on the 2nd run, doesn't it?

chef/cookbooks/ceph/recipes/conf.rb
mmnelemane: @dirkmueller: agreed. Will update to provide the values directly in ceph.conf.erb.

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -66,6 +66,37 @@
   supports restart: true
   action [:enable, :start]
   subscribes :restart, "template[/etc/ceph/ceph.conf]"
+  notifies :run, "execute[trigger_radosgw_metering_setup]", :delayed
+end
+
+execute "trigger_radosgw_metering_setup" do
```
mjura: This looks wrong to me. Why do you use this block to trigger 3 others blocks ?

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -66,6 +66,37 @@
...
+  command "true"
+  notifies :run, "execute[create_radosgw_admin_user]", :delayed
+  notifies :run, "execute[add_radosgw_buckets_caps]", :delayed
+  notifies :run, "execute[add_radosgw_users_caps]", :delayed
+end
+
+# Create admin user and add caps to access buckets and users
+uid = "rgw_admin"
+
+execute "create_radosgw_admin_user" do
```
mjura: On this block chef-client is locking during deployment, maybe radosgw is not ready

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -66,6 +66,37 @@
   supports restart: true
   action [:enable, :start]
   subscribes :restart, "template[/etc/ceph/ceph.conf]"
+  notifies :run, "execute[trigger_radosgw_metering_setup]", :delayed
```
mjura: I think you should notify only "create_radosgw_admin_user" block

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -66,6 +66,37 @@
...
+
+# Create admin user and add caps to access buckets and users
+uid = "rgw_admin"
+
+execute "create_radosgw_admin_user" do
+  command "radosgw-admin user create --display-name=RadosGW_Admin --uid=#{uid}"
+  not_if "out=$(radosgw-admin user info #{uid}) ; [$? != 0] || echo ${out} | grep -q '#{uid}'"
+  retries 3
+  retry_delay 10
+  action :nothing
```
mjura: And after this notify next block
`notifies :run, "execute[add_radosgw_buckets_caps]", :delayed`

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -66,6 +66,37 @@
...
+end
+
+execute "add_radosgw_buckets_caps" do
+  command "radosgw-admin caps add --uid=#{uid} --caps='buckets=*'"
+  only_if "out=$(radosgw-admin user info #{uid}) ; [$? != 0] || echo ${out} | grep -q '#{uid}'"
+  action :nothing
```
mjura: And after this notify next block
`notifies :run, "execute[add_radosgw_users_caps]", :delayed`

mjura:
Doesn't work for me

mjura:
Ok, problem is that `chef/cookbooks/ceph/recipes/radosgw.rb` recipe is run before ceph cluster is created. RadosGW can't store data in RADOS cluster, because OSD'a are not even activated.

mmnelemane:
@mjura: Thanks a lot for the hint. I will update with ceph health check and rework on your comments.

chef/cookbooks/ceph/templates/default/ceph.conf.erb
```diff
@@ -148,8 +148,10 @@
   log file = /var/log/radosgw/radosgw.log
   rgw frontends = civetweb port=<%= @rgw_port %><% if (@rgw_pemfile) %>s ssl_certificate=<%= @rgw_pemfile %><% end %>
   <% unless node[:ceph][:keystone_instance].nil? || node[:ceph][:keystone_instance].empty? -%>
+  rgw s3 auth use keystone = <%= true %>
   rgw keystone url =  <%= @keystone_settings['admin_auth_url'] %>
   rgw keystone admin token = <%= @keystone_settings['admin_token'] %>
+  rgw keystone accepted roles = <%= '"rgw_admin" "Member" "_member_"' %>
```
vuntz: What's the default here? (so I can understand why we change this)

mmnelemane: Default would be "admin", "Member", "_member_". rgw_admin is expected by ceilometer which creates the user to obtain the metrics on ceph objects.

chef/cookbooks/ceph/templates/default/ceph.conf.erb
```diff
@@ -148,8 +148,11 @@
   log file = /var/log/radosgw/radosgw.log
   rgw frontends = civetweb port=<%= @rgw_port %><% if (@rgw_pemfile) %>s ssl_certificate=<%= @rgw_pemfile %><% end %>
   <% unless node[:ceph][:keystone_instance].nil? || node[:ceph][:keystone_instance].empty? -%>
+  rgw s3 auth use keystone = <%= true %>
```
mjura: What is the sense of putting `true` in ruby brackets ?

chef/cookbooks/ceph/templates/default/ceph.conf.erb
```diff
@@ -148,8 +148,11 @@
...
   rgw keystone url =  <%= @keystone_settings['admin_auth_url'] %>
   rgw keystone admin token = <%= @keystone_settings['admin_token'] %>
+  ### default rgw keystone accepted roles : admin, Member, _member_
+  rgw keystone accepted roles = <%= '"rgw_admin" "Member" "_member_"' %>
```
mjura: What is the sense of putting string `"rgw_admin" "Member" "_member_"` into ruby brackets?

mjura: Btw. "rgw_admin" is already hardcoded in `https://github.com/crowbar/crowbar-openstack/pull/88/files#diff-eb55ad1bee5534a01b2bb60eb77a07c4R107` we would like to have it configured only in one place

chef/cookbooks/ceph/templates/default/ceph.conf.erb
mmnelemane: the brackets came from the previous commits which used a variable here. will correct

jsuchome:
This seems to be lost in time. @tserong could you please take a look if it still makes sense?

tserong:
> This seems to be lost in time. @tserong could you please take a look if it still makes sense?

I honestly don't know. The most recent comment from @mjura said that rgw_admin (which this commit adds to "rgw keystone accepted roles") was hardcoded in https://github.com/crowbar/crowbar-openstack/pull/88/files, but that doesn't seem to exist anymore (and `git grep rgw_admin` in a crowbar-openstack checkout shows nothing).

I don't know how ceilometer uses rgw. https://github.com/crowbar/crowbar-ceph/commit/a7429fa seems to have added a keystone user of some kind intended for admin. Is that at all sufficient for this purpose?

jsuchome:
@rhafer, last mentioned commit was yours, so is it relevant to this PR?

rhafer:
> @rhafer, last mentioned commit was yours, so is it relevant to this PR?

Not directly I think. The mentioned commit was needed because we dropped the the admin_token from keystone.

But to be honest I don't really know how relevant this commit still is. At least in its current form it seems to be pretty useless, as we don't create a user that has the "rgw_admin" role anywhere in crowbar.

jdsn:
Closing because the last commit has too many statuses (set by out CI tool). This is blocking rebuild of other PRs. Please reopen for next (force-)push.

PR #22, opened 2015-11-17
Increase limits for ceph-mon and ceph-osd nodes

mjura:
I don't know why we have such low limits for ceph-mon and ceph-osd numbers, but it is wrong.


tboerger:
If you say these values are making sense you get my :+1:

vuntz:
@tserong are you fine with this?

tserong:
Fine by me

vuntz:
+1

PR #26, opened 2015-12-09
Do not pass delaylog option when mounting xfs partitions

vuntz:
This option is deprecated for quite some time, was marked as scheduled
for removal in kernel 3.12 back in 2013-07:
    http://oss.sgi.com/archives/xfs/2013-07/msg00143.html
and was dropped in kernel 3.20 in 2015-02:
    http://oss.sgi.com/archives/xfs/2015-02/msg00306.html

Note that we have 3.12 in SLES 12 SP1 and 4.1.13 in Leap.


dirkmueller:
:+1:

tboerger:
:+1:

aplanas:
+1

PR #27, opened 2015-12-09
Allow deploying ceph roles on Leap

vuntz:


dirkmueller:
the commit message says "leap" but the change actually allows all of opensuse.. which one did you want to do ?

vuntz:
@dirkmueller I'm kind of pretending that openSUSE <= 13.2 doesn't exist, to be honest ;-) I can make this more specific if wanted (42.1).

Would >= 42.1 be appropriate? This would also enable Tumbleweed as a sideeffect (in case people want to have fun)

mjura:
I think that we should stay with 42.1. I'm not sure about others releases.

tboerger:
I think we can enable openSUSE in general and give other versions a shot :)

tboerger:
:+1:

aplanas:
+1

mjura:
> I think we can enable openSUSE in general and give other versions a shot :)

But we will support only openSUSE Leap nodes and for SLES we are supporting 12.1. So the question is if we want to be so liberal why we require specific version for SLES.

tboerger:
> But we will support only openSUSE Leap nodes and for SLES we are supporting 12.1. So the question is if we want to be so liberal why we require specific version for SLES.

Currently we don't support any openSUSE install officially. Beside that it's much more difficult to get matching versions approved on SLES than on openSUSE. If we get in trouble with some openSUSE versions we can still restrict it. Those are just my 2 cents.

PR #29, opened 2015-12-16
radosgw: Limit ceph.conf perms when it has keystone admin token (bsc#958470)

tserong:
/etc/ceph/ceph.conf is usually chmod 644, user and group root.  When
deploying radosgw with keystone, the keystone admin token ends up in
ceph.conf, so in this case it's chmod 640, user root, group www (the
latter being necessary so the radosgw process can actually still read
the file).

### DIFF
--- a/chef/cookbooks/ceph/recipes/conf.rb
+++ b/chef/cookbooks/ceph/recipes/conf.rb
@@ -118,9 +118,55 @@ template "/etc/ceph/ceph.conf" do
     is_rgw: is_rgw,
     rgw_port: rgw_port,
     rgw_pemfile: rgw_pemfile,
-    keystone_settings: keystone_settings
+    keystone_settings: node["platform_family"] == "suse" ? {} : keystone_settings
   )
   mode "0644"
+end
+
+if is_rgw && node["platform_family"] == "suse"
+  # We create a separate ceph.conf file for use only by the radosgw daemon,
+  # and an override systemd unit file which makes radosgw use this config file,
+  # so we can limit its permissions as it may include the keystone admin token.
+  #
+  # It's not sufficient to just set the main ceph.conf to 0640, because other
+  # services (glance/cinder/nova) may need to read ceph.conf.
+  #
+  # Note that ceph.conf still includes a radosgw section, because that's necessary
+  # for the ceph-radosgw-prestart.sh script to read.  ceph.conf.radosgw is thus
+  # almost identical to ceph.conf, but it also includes the keystone settings
+  # if applicable.
+  #
+  # (The override systemd unit file is created in radosgw.rb, rather than here,
+  # because this needs to happen *after* the radosgw pacakge is installed, or
+  # the main systemd unit file doesn't exist yet)
+  #
+  # The separate, radosgw specific config file must be named ceph.conf.radosgw,
+  # or, at least, not look like it's named "*.conf", because otherwise it may be
+  # picked up erroneously when the ceph command line tools are scanning /etc/ceph
+  # for viable config files.  We *only* want this to be used when the radosgw
+  # daemon runs, not for anything else (an earlier attempt, where the file was
+  # named "ceph-radosgw.conf", made `ceph-disk activate` think the cluster was
+  # named "ceph-radosgw", and then it couldn't find a key for that cluster name...)
+  #
+  template "/etc/ceph/ceph.conf.radosgw" do
+    source "ceph.conf.erb"
+    variables(
+      mon_initial: mon_init,
+      mon_addresses: mon_addr,
+      pool_size: rep_num,
+      pool_pg_num: pg_num,
+      osd_nodes_count: osd_nodes.length,
+      public_network: node["ceph"]["config"]["public-network"],
+      cluster_network: node["ceph"]["config"]["cluster-network"],
+      is_rgw: is_rgw,
+      rgw_port: rgw_port,
+      rgw_pemfile: rgw_pemfile,
+      keystone_settings: keystone_settings
+    )
+    owner "root"
+    group node["ceph"]["radosgw"]["group"]
+    mode "0640"
+  end
 end
 
 # Need salt minion on osd and mon nodes to hook up to calamari.  Note: in one
--- a/chef/cookbooks/ceph/recipes/radosgw.rb
+++ b/chef/cookbooks/ceph/recipes/radosgw.rb
@@ -42,11 +42,40 @@ unless node[:ceph][:keystone_instance].nil? || node[:ceph][:keystone_instance].e
   include_recipe "ceph::radosgw_keystone"
 end
 
+rgw_conf = "ceph.conf"
+
+if node["platform_family"] == "suse"
+  rgw_conf = "ceph.conf.radosgw"
+  # When generating the override systemd unit file, we have to disable the
+  # radosgw service so that it gets re-enabled later, thus picking up the
+  # override unit file (if we didn't do this, a host that was already
+  # running radosgw and was then upgraded wouldn't pick up the new unit file).
+  # (see also the comments about this in conf.rb)
+  # The first sed strips any existing --conf option out of the ExecStart line
+  # (this was present in Hammer, but is no longer present in Jewel).  The
+  # second sed tells radosgw to explicitly use our radosgw specific conf file.
+  bash "generating override ceph-radosgw systemd unit file" do
+    code <<-EOH
+      sed -e 's%^\\(ExecStart=.*\\)\\(--conf [^ ]*\\)\\(.*\\)%\\1\\3%' \
+        /usr/lib/systemd/system/ceph-radosgw@.service | \
+      sed -e 's%^\\(ExecStart=.*\\)%\\1 --conf /etc/ceph/ceph.conf.radosgw%' \
+         > /etc/systemd/system/ceph-radosgw@.service
+      systemctl daemon-reload
+      systemctl disable #{node["ceph"]["radosgw"]["service_name"]}
+    EOH
+    not_if do
+      File.exist?("/etc/systemd/system/ceph-radosgw@.service") &&
+        File.mtime("/etc/systemd/system/ceph-radosgw@.service") >
+          File.mtime("/usr/lib/systemd/system/ceph-radosgw@.service")
+    end
+  end
+end
+
 service "radosgw" do
   service_name node["ceph"]["radosgw"]["service_name"]
   supports restart: true
   action [:enable, :start]
-  subscribes :restart, "template[/etc/ceph/ceph.conf]"
+  subscribes :restart, "template[/etc/ceph/#{rgw_conf}]"
 end
 
 # In the systemd case, need extra targets enabled
--- a/chef/cookbooks/ceph/templates/default/ceph.conf.erb
+++ b/chef/cookbooks/ceph/templates/default/ceph.conf.erb
@@ -135,7 +135,7 @@
   host = <%= node['hostname'] %>
   keyring = /etc/ceph/ceph.client.rgw.<%= node['hostname'] %>.keyring
   rgw frontends = civetweb port=<%= @rgw_port %><% if (@rgw_pemfile) %> ssl_certificate=<%= @rgw_pemfile %><% end %>
-  <% unless node[:ceph][:keystone_instance].nil? || node[:ceph][:keystone_instance].empty? -%>
+  <% unless node[:ceph][:keystone_instance].nil? || node[:ceph][:keystone_instance].empty? || @keystone_settings.empty? -%>
   rgw keystone url =  <%= @keystone_settings['admin_auth_url'] %>
   rgw keystone admin user = <%= @keystone_settings['service_user'] %>
   rgw keystone admin tenant = <%= @keystone_settings['service_tenant'] %>


tserong:
Hrm. Why is gating red? Something about tempest not being able to upload an image? Not sure how that relates to this change... :-/

mjura:
Problem with this change will be with Glance, Cinder and Nova  ceph integration. All this services run with specific users and they need to have access to read ceph.conf. Right now glance, cinder and nova users are not members of www group.

tserong:
Ah. OK. So removing world read basically breaks everything :-/

tboerger:
Codewise :+1: but needs further investigation :)

vuntz:
Maybe we should simply create another ceph.conf file just for radosgw?

mjura:
> Maybe we should simply create another ceph.conf file just for radosgw?

This is good idea, @tserong can you check if it will work?

tserong:
> Maybe we should simply create another ceph.conf file just for radosgw?
> 
> This is good idea, @tserong can you check if it will work?

Wait a minute, that's what this change does (or should do), i.e. it only applies those perms on the node that's running radosgw. Or do you mean having a separate ceph.conf file (ceph-radosgw.conf?) which is only used by the radosgw daemon?

Are glance/cinder/nova allowed to run on a node that's also running radosgw?

vuntz:
> Or do you mean having a separate ceph.conf file (ceph-radosgw.conf?) which is only used by the radosgw daemon?

Yes, separate ceph.conf file.

> Are glance/cinder/nova allowed to run on a node that's also running radosgw?

At the moment, I think there's no such restriction, indeed.

tserong:
> Or do you mean having a separate ceph.conf file (ceph-radosgw.conf?) which is only used by the radosgw daemon?
> 
> Yes, separate ceph.conf file.

This might get weird.

The ceph-radosgw systemd unit file as shipped in the ceph package specifies the config file for the daemon to use:

```
[Service]
EnvironmentFile=-/etc/sysconfig/ceph
Environment=CLUSTER=ceph
ExecStart=/usr/bin/radosgw -f --conf /etc/ceph/${CLUSTER}.conf --name client.radosgw.%i
```

That's parameterized to allow running clusters with names other that "ceph". ceph-osd and ceph-mon do the same thing. At first glance it would seem that we could abuse this feature and set set `CLUSTER=something-else` in `/etc/sysconfig/ceph`, which would cause radosgw to use /etc/ceph/something-else.conf. But it would also cause ceph-mon and ceph-osd to run with `--cluster something-else`, i.e. suddenly the cluster wouldn't be named "ceph" anymore as far as all the daemons were concerned. All nodes would need the alternate config file. In theory that might actually work if you started out your deployment that way, but if you had an existing deployment, then added radosgw, and suddenly the cluster name changed, well... It might suddenly become a smoking wreck. Not that I've tried this, but it sounds dangerous, so let's not ;-)

A safer alternative might be to have the barclamp create an override unit file in `/etc/systemd/system` which would be used instead of the regular ceph-radosgw service file, and could specify an alternate config file without changing anything for any of the other ceph daemons. The catch here is that we'd need to make sure this unit file remained in sync with what was shipped with ceph.

vuntz:
@tserong I wouldn't change the name of the ceph cluster, that's quite bad :-)

Can we possibly pass a second `--conf` option? If yes, maybe we should do that with an override of the service file.

tserong:
You can't pass multiple config files (well, strictly speaking you can, but it takes the first one it can find and ignores the rest). We'd have to have two almost identical ceph config files, with the second one used by radosgw including the extra keystone attributes. I think I can make this work...

tserong:
Huh. Guess not.

tserong:
Actually, can someone who knows what they're doing please have a look at that last gating failure? Unless I'm misreading it, the problem appears to be:

```
======================================================================
FAIL: tearDownClass (tempest.scenario.test_server_basic_ops.TestServerBasicOps)
----------------------------------------------------------------------
Traceback (most recent call last):
_StringException: Traceback (most recent call last):
  File "/var/lib/openstack-tempest-test/tempest/test.py", line 312, in tearDownClass
    six.reraise(etype, value, trace)
  File "/var/lib/openstack-tempest-test/tempest/test.py", line 295, in tearDownClass
    teardown()
  File "/var/lib/openstack-tempest-test/tempest/test.py", line 530, in clear_credentials
    cls._creds_provider.clear_creds()
  File "/var/lib/openstack-tempest-test/tempest/common/dynamic_creds.py", line 327, in clear_creds
    self._clear_isolated_net_resources()
  File "/var/lib/openstack-tempest-test/tempest/common/dynamic_creds.py", line 318, in _clear_isolated_net_resources
    creds.subnet['name'])
  File "/var/lib/openstack-tempest-test/tempest/common/dynamic_creds.py", line 269, in _clear_isolated_subnet
    client.delete_subnet(subnet_id)
  File "/var/lib/openstack-tempest-test/tempest/services/network/json/subnets_client.py", line 34, in delete_subnet
    return self.delete_resource(uri)
  File "/var/lib/openstack-tempest-test/tempest/services/network/json/base.py", line 40, in delete_resource
    resp, body = self.delete(req_uri)
  File "/usr/lib/python2.7/site-packages/tempest_lib/common/rest_client.py", line 290, in delete
    return self.request('DELETE', url, extra_headers, headers, body)
  File "/usr/lib/python2.7/site-packages/tempest_lib/common/rest_client.py", line 639, in request
    resp, resp_body)
  File "/usr/lib/python2.7/site-packages/tempest_lib/common/rest_client.py", line 702, in _error_checker
    raise exceptions.Conflict(resp_body, resp=resp)
Conflict: An object with that identifier already exists
Details: {u'message': u'Unable to complete operation on subnet a80e4f2f-dc77-445e-98b9-2ef2c903bc08. One or more ports have an IP allocation from this subnet.', u'type': u'SubnetInUse', u'detail': u''}
```

I've no idea what would cause that :-(

vuntz:
@tserong gating is broken when ceph is used at the moment (at least two tests are failing). In your case, you have two more tests failing, but they look like errors that can happen from time to time, not bugs.

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -89,11 +89,58 @@
...
+
+  # When generating the override systemd unit file, we have to disable the
+  # radosgw service so that it gets re-enabled later, thus picking up the
+  # override unit file (if we didn't do this, a host that was already
+  # running radosgw and was then upgraded wouldn't pick up the new unit file).
+  bash "generating override ceph-radosgw systemd unit file" do
+    code <<-EOH
+      sed -e 's/\\(ExecStart=.*\\)\\(\\.conf.*\\)/\\1-radosgw\\2/' \
+        /usr/lib/systemd/system/ceph-radosgw@.service > /etc/systemd/system/ceph-radosgw@.service
+      systemctl disable #{node["ceph"]["radosgw"]["service_name"]}
```
vuntz: Wow, looks evil :-)

I don't think you should disable the service, but you should do "systemctl daemon-reload".

tserong:
@vuntz, ah OK, thanks.

In that case, what I _can_ say about my last commit based on what's in the supportconfig tarballs from that gating run, is that the right stuff ends up in the alternate config file and the radosgw daemon is using it.

chef/cookbooks/ceph/recipes/conf.rb
tserong: `systemctl daemon-reload` doesn't cut it, unfortunately. The (possibly bizarre edge) case I'm trying to handle here is where an older system, prior to this change, already has radosgw enabled. In this case, the original enablement of radosgw produces the following symlink to the stock unit file:

`/etc/systemd/system/ceph.target.wants/ceph-radosgw@d52-54-00-49-17-2a.service -> /usr/lib/systemd/system/ceph-radosgw@.service`

Doing `systemctl daemon-reload` just reloads the contents of existing unit files, it won't update symlinks for enabled services to new override unit files that have appeared.

If I disable the service (which removes that symlink), the `service` block at https://github.com/crowbar/crowbar-ceph/blob/master/chef/cookbooks/ceph/recipes/radosgw.rb#L64 subsequently triggers an enable, which finds the newly created override unit file and sets up the correct symlink:

`/etc/systemd/system/ceph.target.wants/ceph-radosgw@d52-54-00-49-17-2a.service -> /etc/systemd/system/ceph-radosgw@.service`

And yes, I agree this is probably evil. It's getting slightly too close to "works by unexplained miracle" for my liking, but I haven't found a cleaner solution.

tserong: Oh, wait, I should probably _also_ do `systemctl daemon-reload`, because the `sed` can potentially change the override unit file if the stock unit file has changed in the meantime... Shall fix.

tserong:
(rebased and combined a couple of fixup commits)

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -89,11 +89,57 @@
...
+  # override unit file (if we didn't do this, a host that was already
+  # running radosgw and was then upgraded wouldn't pick up the new unit file).
+  bash "generating override ceph-radosgw systemd unit file" do
+    code <<-EOH
+      sed -e 's/\\(ExecStart=.*\\)\\(\\.conf.*\\)/\\1-radosgw\\2/' \
+        /usr/lib/systemd/system/ceph-radosgw@.service > /etc/systemd/system/ceph-radosgw@.service
+      systemctl daemon-reload
+      systemctl disable #{node["ceph"]["radosgw"]["service_name"]}
+    EOH
+  end
```
vuntz: Sorry for the very late review... Shouldn't there be something like a `not_if { File.exist? "/etc/systemd/system/ceph-radosgw@.service" }`. It feels to me like right now, you are doing a daemon-reload every 15 minutes

tserong: You may be right - leave it with me (I'll need a couple of days due to $other_stuff, sorry :-/)

tserong:
@vuntz, yeah, it was regenerating that file and re-running daemon-reload every fifteen minutes. I've added a `not_if` so that it'll only activate if the file doesn't exist, or exists and is older than the regular unit file.

tserong:
...and now I've made hound stop barking too ;-)

vuntz:
@tserong hrm, the gating tells me it doesn't like your change:

```
11:26:11 [2016-03-18T10:26:09+00:00] INFO: Processing service[radosgw] action start (ceph::radosgw line 66)
11:26:11 
11:26:11 ================================================================================
11:26:11 Error executing action `start` on resource 'service[radosgw]'
11:26:11 ================================================================================
11:26:11 
11:26:11 Chef::Exceptions::Exec
11:26:11 ----------------------
11:26:11 /bin/systemctl start ceph-radosgw@d52-54-77-77-77-01 returned 1, expected 0
11:26:11 
11:26:11 Resource Declaration:
11:26:11 ---------------------
11:26:11 # In /var/chef/cache/cookbooks/ceph/recipes/radosgw.rb
11:26:11 
11:26:11  66: service "radosgw" do
11:26:11  67:   service_name node["ceph"]["radosgw"]["service_name"]
11:26:11  68:   supports restart: true
11:26:11  69:   action [:enable, :start]
11:26:11  70:   subscribes :restart, "template[/etc/ceph/#{rgw_conf}]"
11:26:11  71: end
11:26:11  72: 
11:26:11 
11:26:11 Compiled Resource:
11:26:11 ------------------
11:26:11 # Declared in /var/chef/cache/cookbooks/ceph/recipes/radosgw.rb:66:in `from_file'
11:26:11 
11:26:11 service("radosgw") do
11:26:11   action [:enable, :start]
11:26:11   updated true
11:26:11   supports {:restart=>true}
11:26:11   retries 0
11:26:11   retry_delay 2
11:26:11   service_name "ceph-radosgw@d52-54-77-77-77-01"
11:26:11   enabled true
11:26:11   pattern "radosgw"
11:26:11   startup_type :automatic
11:26:11   cookbook_name "ceph"
11:26:11   recipe_name "radosgw"
11:26:11 end
```

tserong:
> @tserong hrm, the gating tells me it doesn't like your change:

Well, that sucks. I can't seem to figure out why `systemctl start ceph-radosgw@d52-54-77-77-77-01` failed from the supportconfig tarballs. It just failed. I'll have to try to reproduce here...

tserong:
/me facepalms

The override unit file was empty, because it was created before the ceph-radosgw package was installed, so the regular unit file didn't exist yet when I was sedding it. Naturally I didn't hit this, because I did my testing on a VM that already had everything installed - I hadn't tried a from-scratch deploy with my change applied.

tserong:
OK, so, radosgw is now actually starting properly since my last change, but gating gave me:

```
07:15:01 ======================================================================
07:15:01 FAIL: test10_ValidateOSDStat (testcloud_sanity.TestSanity)
07:15:01 ----------------------------------------------------------------------
07:15:01 Traceback (most recent call last):
07:15:01   File "/root/qa-automation/testsuites/testcloud_sanity.py", line 125, in test10_ValidateOSDStat
07:15:01     assert(expStr in osd_stat),"osd stat validation failed"
07:15:01 AssertionError: osd stat validation failed
07:15:01 -------------------- >> begin captured logging << --------------------
07:15:01 testcloud_sanity: INFO: ++++++starting test10_ValidateOSDStat ++++++
07:15:01 utils.launch: DEBUG: executing the command ssh d52-54-77-77-77-01 ceph osd stat
07:15:01 utils.monitoring: INFO: the osd stat is osdmap e32: 3 osds: 3 up, 3 in
07:15:01 --------------------- >> end captured logging << ---------------------
07:15:01 
07:15:01 ----------------------------------------------------------------------
07:15:01 Ran 15 tests in 141.207s
07:15:01 
07:15:01 FAILED (failures=1)
```

I suspect the test is expecting four OSDs, not three...

tserong:
@vuntz, @mjura, have either of you seen a gating failure like the above before?

tserong:
OK, fixed that gating failure, once I finally figured out WTF was going on (please bear in mind my commit message on https://github.com/crowbar/crowbar-ceph/pull/29/commits/c0afdaf1b44c02113e297dbe0e4ad70c9f96ad21 when reviewing this)

AbelNavarro:
I don't quite like this workaround to get proper access rights to `ceph.conf`, but I don't think there's a better approach right now. It seems it has been [asked in the past](http://lists.ceph.com/pipermail/ceph-users-ceph.com/2013-March/000372.html) to have a separate file for authentication. Also our [competitors](https://content.mirantis.com/rs/451-RBY-185/images/Mirantis-Technical-Bulletin-S3-API-Keystone-integration-in-Ceph-RADOS-Gateway.pdf) don't even take care of that.

So, :+1: for me.

mjura:
Looks ok, I think that we can merge it +1

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -89,11 +89,57 @@
...
 
+if is_rgw && node["platform_family"] == "suse"
+  # We create a separate ceph.conf file for use only by the radosgw daemon,
+  # and an override systemd unit file which makes radosgw use this config file,
+  # so we can limit its permissions as it may include the keystone admin token.
+  #
+  # It's not sufficient to just set the main ceph.conf to 0640, because other
+  # services (glance/cinder/nova) may need to read ceph.conf.
+  #
+  # Note that ceph.conf still includes a radosgw section, because that's necssary
```
nicolasbock: /necssary/necessary/

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -61,11 +61,40 @@
...
+  # radosgw service so that it gets re-enabled later, thus picking up the
+  # override unit file (if we didn't do this, a host that was already
+  # running radosgw and was then upgraded wouldn't pick up the new unit file).
+  # (see also the comments about this in conf.rb)
+  # The first sed strips any existing --conf option out of the ExecStart line
+  # (this was present in Hammer, but is no longer present in Jewel).  The
+  # second sed tells radosgw to explicitly use our radosgw specific conf file.
+  bash "generating override ceph-radosgw systemd unit file" do
+    code <<-EOH
+      sed -e 's%^\\(ExecStart=.*\\)\\(--conf [^ ]*\\)\\(.*\\)%\\1\\3%' \
```
nicolasbock, line 59: You could also add the `--in-place` option which would shorten this a bit.

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -61,11 +61,40 @@
...
+        /usr/lib/systemd/system/ceph-radosgw@.service | \
+      sed -e 's%^\\(ExecStart=.*\\)%\\1 --conf /etc/ceph/ceph.conf.radosgw%' \
```
nicolasbock, line 61: Couldn't you combine the two `sed` expressions?
`s%^\\(ExecStart=.*\\)--conf +[^ ]*\\(.*\\)%\1--conf /usr/..../ceph-radosgw@.service\3%`

nicolasbock:
+1, but please have a look at my comments.

chef/cookbooks/ceph/recipes/radosgw.rb
tserong, line 61: You may well be right (and also with the `--in-place` option). I spent some time testing variants of this, but I honestly can't remember if I tried combining the two or not. Speaking purely from a selfish point of view, I'd rather it just go in as-is if you don't have a strong objection, because it'll save me a couple of rounds of further testing ;-)

tserong:
Rebased

AbelNavarro:
time to merge this?

nicolasbock (approved):
Still +1

tserong:
Am I allowed to press the "merge" button?

PR #30, opened 2015-12-17
Fix for client LWRP whent get_caps() returns empty hash (bsc#959567)

mjura:
https://bugzilla.suse.com/show_bug.cgi?id=959567


aplanas:
+1

vuntz:
+1

vuntz:
The failure in gating is only related to the volume_boot_pattern tests in tempest that are known to be failing right now, so merging.

PR #34, opened 2016-01-18
Let the validation error strings are translatable.

jsuchome:
This is just one string of many (the one I've met in https://github.com/crowbar/crowbar-openstack/pull/201) and I think others should be moved to locale file as well.

Or is there any reason we do not have it? @tserong @mjura

### DIFF
--- a/crowbar_framework/config/locales/ceph/en.yml
+++ b/crowbar_framework/config/locales/ceph/en.yml
@@ -43,3 +43,10 @@ en:
             enabled: 'Protocol'
             certfile: 'SSL Certificate File'
             keyfile: 'SSL (Private) Key File'
+      validation:
+        disk_missing: 'Nodes %{nodes} for ceph-osd role are missing at least one unclaimed disk larger than %{size}GB.'
+        mon_role_missing: 'Nodes in cluster must have same roles: node %{node} is missing ceph-mon role.'
+        osd_role_missing: 'Nodes in cluster must have same roles: node %{node} is missing ceph-osd role.'
+        osd_removal: 'The ceph-osd role cannot be removed from a node %{node}.'
+        ses_repos: 'The SUSE Enterprise Storage repositories have not been setup.'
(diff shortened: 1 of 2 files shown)


aplanas:
+1

tboerger:
:+1:

crowbar_framework/config/locales/ceph/en.yml
```diff
@@ -43,3 +43,5 @@ en:
             enabled: 'Protocol'
             certfile: 'SSL Certificate File'
             keyfile: 'SSL (Private) Key File'
+      validation:
+        ses_repos: 'The SUSE Enterprise Storage repositories have not been setup.'
```
vuntz, line 51: So, it's not introduced by your commit, but I think the error string should be made vendor-neutral. Something like "Ceph is not available in the repositories.", and then we would patch the string in the package to mention SES.

vuntz:
+1; I'm happy to do a followup PR for my comment.

jsuchome:
I've added more strings from ceph_service.rb

tserong:
LGTM

nkrinner:
+1

crowbar_framework/config/locales/ceph/en.yml
```diff
@@ -43,3 +43,11 @@ en:
             enabled: 'Protocol'
             certfile: 'SSL Certificate File'
             keyfile: 'SSL (Private) Key File'
+      validation:
+        disk_missing: "Nodes %{nodes} for ceph-osd role are missing at least one unclaimed disk larger than %{size}GB."
```
tboerger: Please use the single quotes like it's done for all other keys.

tboerger:
After making the quotes more unified you get my :+1:

vuntz:
+1

dirkmueller:
This commit apparently causes https://bugzilla.suse.com/show_bug.cgi?id=967255

PR #35, opened 2016-02-05
Fix possible deadlock because of raise command, (bsc#965324)

mjura:


vuntz:
But won't the `ceph-authtool` call be wrong because it will be adding an empty key?

mjura:
@vuntz : I adjusted this changes better for you :)

tserong:
+1

vuntz:
Similar comment to the other PR: would it be something that can be resolved by moving the code to convergence phase?

mjura:
> Similar comment to the other PR: would it be something that can be resolved by moving the code to convergence phase?

I'm not sure if we should make exception for this code here. Before this change old `if` condition comes upstream cookbook which was re-worked, few times. Right new check looks more readable.

As always I'm open for any suggestion on how we should modify this bit of code.

vuntz:
Same here, let's not lose too much time on this now. I documented the issue on https://etherpad.nue.suse.com/p/cloud-orchestration

PR #38, opened 2016-02-19
Workaround Ceph hanging QEMU on volume attach (bsc#967509)

dirkmueller:


toabctl:
+1 (also tested it, works).

vuntz:
So that doesn't help us for external ceph clusters. I'm wondering if it's worth working around it here in that case.

dirkmueller:
Yeah, but given the frequency and the impact of the bug I'd consider a workaround necessary until maintenance update is released - that also fixes the external ceph case.

vuntz:
Hrm, okay. As long as we track that this needs to be reverted...

PR #39, opened 2016-03-10
Set rake to version to 10.3.2

rsalevsky:
This change is needed due to the incompatibility with rake 11.0.0.


tboerger:
:+1:

dirkmueller:
:+1:

tboerger:
I think we should revert that as it seems to be not needed anymore.

PR #41, opened 2016-03-11
Run syntaxcheck with warnings enabled

dirkmueller:
That gives some useful hints on what could be improved
as well.


tboerger:
Yeah why not.
:+1:

mjura:
:+1:

PR #44, opened 2016-03-23
Revert "Workaround Ceph hanging QEMU on volume attach (bsc#967509)"

toabctl:
This reverts commit c7aceac83159856ff14ce414f2fd6fa795dc2e69. It's
no longer needed because it's fixed in Ceph.


dirkmueller:
Assuming that gating passes, +1

vuntz:
+1

PR #45, opened 2016-04-04
Added rubocop config to map to hound config

tboerger:


tboerger:
Merge?

toabctl:
+1 (but are we skipping our 2x +1 rules now?)

tboerger:
We are not skipping it but exactly the same pr have been accepted for the other repos

PR #46, opened 2016-04-05
Support deploying Ceph Jewel release (bsc#970510)

tserong:
The Ceph Jewel release runs the various ceph daemons as an unprivileged user and group (ceph:ceph), whereas previous releases ran everything as root. There's a bunch of stuff that needs changing in the barclamp to support this. I believe I've got most of it in this PR, but still need to test further.

Also, the changes here effectively _only_ work with Jewel (they rely on the ceph package having created the appropriate user accounts automatically), so if you use this code with an older Ceph release, it will almost certainly fail to deploy. I may be better off trying to make these changes conditionally depending on what version of Ceph is available, but I don't actually know how to check that from within the Chef recipes.

### DIFF
--- a/chef/cookbooks/ceph/recipes/conf.rb
+++ b/chef/cookbooks/ceph/recipes/conf.rb
@@ -9,6 +9,8 @@ mon_nodes.each do |monitor|
     mon_init << monitor.name.split(".")[0]
 end
 
+# These directories should all be created by the package, maybe we can
+# just remove this stuff...?  (same with the directories created in osd.rb)
 directory "/etc/ceph" do
   owner "root"
   group "root"
@@ -17,16 +19,16 @@ directory "/etc/ceph" do
 end
 
 directory "/var/run/ceph" do
-  owner "root"
-  group "root"
-  mode "0755"
+  owner "ceph"
+  group "ceph"
+  mode "0770"
   action :create
 end
 
 directory "/var/log/ceph" do
-  owner "root"
-  group "root"
-  mode "0755"
+  owner "ceph"
+  group "ceph"
+  mode "3770"
   action :create
 end
 
--- a/chef/cookbooks/ceph/recipes/conf.rb
+++ b/chef/cookbooks/ceph/recipes/conf.rb
@@ -9,27 +9,29 @@ mon_nodes.each do |monitor|
     mon_init << monitor.name.split(".")[0]
 end
 
-# These directories should all be created by the package, maybe we can
-# just remove this stuff...?  (same with the directories created in osd.rb)
-directory "/etc/ceph" do
-  owner "root"
-  group "root"
-  mode "0755"
-  action :create
-end
+unless node["platform_family"] == "suse"
+  # These directories are all created by the ceph packages on SUSE distros.
+  # TODO: Check if this is true for other distros (it probably is)
+  directory "/etc/ceph" do
+    owner "root"
+    group "root"
+    mode "0755"
+    action :create
+  end
 
-directory "/var/run/ceph" do
-  owner "ceph"
-  group "ceph"
-  mode "0770"
-  action :create
-end
+  directory "/var/run/ceph" do
+    owner "ceph"
+    group "ceph"
+    mode "0770"
+    action :create
+  end
 
-directory "/var/log/ceph" do
-  owner "ceph"
-  group "ceph"
-  mode "3770"
-  action :create
+  directory "/var/log/ceph" do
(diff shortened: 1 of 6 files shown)


tserong:
radosgw deployment is presently broken

tserong:
The gating failure is because the ceph user doesn't exist (this is never going to succeed anyway until we've got the CI system using SES 3 with Jewel)

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -17,16 +19,16 @@
...
   action :create
 end
 
 directory "/var/log/ceph" do
-  owner "root"
-  group "root"
-  mode "0755"
+  owner "ceph"
+  group "ceph"
+  mode "3770"
```
AbelNavarro: is that **sticky bit** and **setguid** necessary?

tserong: I don't know why that's necessary offhand, but it matches what's in the ceph spec file:

`%attr(3770,ceph,ceph) %dir %{_localstatedir}/log/ceph/`

AbelNavarro:
triggered rebuild.

tserong:
From my perspective this is good to go, once https://github.com/SUSE-Cloud/automation/pull/991 is merged and the SES3 repos are available. Unless these changes need to be somehow made conditional on the version of Ceph that's available to install...? To be clear, if this PR is accepted as-is, Ceph >= Jewel becomes a requirement, i.e. it won't correctly deploy older Ceph versions.

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -9,6 +9,8 @@
     mon_init << monitor.name.split(".")[0]
 end
 
+# These directories should all be created by the package, maybe we can
+# just remove this stuff...?  (same with the directories created in osd.rb)
```
vuntz: It's unclear if they're needed for other distros, so one approach would be to only create them for non-SUSE platform families.

chef/cookbooks/ceph/recipes/radosgw.rb
```diff
@@ -68,6 +49,16 @@
   subscribes :restart, "template[/etc/ceph/ceph.conf]"
 end
 
+# In the systemd case, need extra targets enabled
+service "ceph-radosgw.target" do
+  action :enable
+  only_if { File.exist?("/usr/lib/systemd/system/ceph-radosgw.target") }
```
vuntz, line 55: Why the `only_if`?

chef/cookbooks/ceph/recipes/mon.rb
```diff
@@ -91,7 +91,7 @@
   end
 
   execute "ceph-mon mkfs" do
-    command "ceph-mon --mkfs -i #{node['hostname']} --keyring '#{keyring}'"
+    command "chown ceph:ceph #{keyring} ; ceph-mon --mkfs -i #{node['hostname']} --keyring '#{keyring}' --setuser ceph --setgroup ceph"
```
vuntz, line 94: Is the `chown` required only when migrating from a previous release, or all the time? I find this slightly suboptimal, but I don't have a better proposal right now.

chef/cookbooks/ceph/recipes/radosgw.rb
tserong, line 55: mon.rb and osd.rb use `if service_type == "systemd"`, and `service_type` comes from `node["ceph"]["mon"]["init_style"]` or `node["ceph"]["osd"]["init_style"]`. There's no such equivalent attribute for rgw, so I went with `only_if` on the unit file instead.

chef/cookbooks/ceph/recipes/mon.rb
tserong, line 94: The keyring needs to be owned by the ceph user so the daemons can read it, but (IIRC) because the earlier `ceph-authtool` invocations are being run by chef as root, the keyring ends up with root ownership by default.

chef/cookbooks/ceph/recipes/radosgw.rb
vuntz, line 55: Hrm, okay. I guess another way is to just put that inside a platform family check for "suse". But no big deal.

vuntz:
+1, but needs to be merged with https://github.com/SUSE-Cloud/automation/pull/991 and https://github.com/crowbar/crowbar-core/pull/483

PR #47, opened 2016-04-06
Handle down nodes when searching for MONs and OSDs (bsc#972001)

tserong:
Searching for OSDs using the query "run_list:role[ceph-osd]" works
fine if all the nodes are up, but if any OSD nodes are down, they're not
returned by this search (somehow the ceph-osd role disappears from the
role run list when the node goes down, and reappears when it comes back
up).  This can lead to a case where a node which is a ceph client hits a
ZeroDivisionError applying the ceph recipe if it comes up before any
OSDs are available (OSD count is zero in this case, at least
transiently).

This commit changes the search to query "run_list_map:ceph_osd", which
seems to work fine regardless of whether nodes are up or down.

### DIFF
--- a/chef/cookbooks/ceph/recipes/conf.rb
+++ b/chef/cookbooks/ceph/recipes/conf.rb
@@ -63,37 +63,47 @@ else
   num_osds = node["ceph"]["config"]["osds_in_total"]
 end
 
-# There'll always be at least the default rbd pool (which Ceph always creates
-# with 64 PGs, irrespective of what "osd pool default pg num" is set to)
-expected_pools = 1
+case num_osds
+when 0
+  # This can only happen if the search in get_osd_nodes() doesn't find any OSDs, which
+  # should be impossible.  But if it *does* fail, pg_num needs to be set to something,
+  # so let's run with the default of 8 (see "osd pool default pg num" at
+  # http://docs.ceph.com/docs/master/rados/configuration/pool-pg-config-ref/)
+  Chef::Log.warn("Ceph recipe invoked but there are no OSDs!  Defaulting to pg_num = 8")
+  pg_num = 8
+else
+  # There'll always be at least the default rbd pool (which Ceph always creates
+  # with 64 PGs, irrespective of what "osd pool default pg num" is set to)
+  expected_pools = 1
 
-rgw_host = search(:node, "roles:ceph-radosgw")
-# RGW will use up to 14 pools (try "RGW only" with http://ceph.com/pgcalc/)
-# These are not all created immediately though.  Six will be created when
-# the radosgw daemon starts for the first time.  The swift pool(s) will be
-# created when you create a swift subuser for the first time, and the usage
-# pool will only be created if the usage log is explicitly enabled.  Here,
-# to be conservative, we're assuming the full 14 pools will be used for RGW
-# deployments.
-expected_pools += 14 unless rgw_host.empty?
+  rgw_host = search(:node, "roles:ceph-radosgw")
+  # RGW will use up to 14 pools (try "RGW only" with http://ceph.com/pgcalc/)
+  # These are not all created immediately though.  Six will be created when
+  # the radosgw daemon starts for the first time.  The swift pool(s) will be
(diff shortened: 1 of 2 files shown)


tserong:
Anyone fancy reviewing this? @mjura @vuntz? Or someone else, if I'm doing weird things with chef searches?

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -62,6 +62,9 @@
 # calculate pg_num based on documentation
 # http://ceph.com/docs/master/rados/operations/placement-groups/
 case num_osds
+when 0
+  Chef::Log.warn("Ceph recipe invoked but there are no OSDs!")
+  return
```
dirkmueller: hmm, do you really want to return here?

tserong: With the change to the search criteria, it shouldn't be possible to hit the zero case anymore, but if it somehow _does_ come up zero, I thought bailing out somewhat cleanly was less worse than hitting a divide by zero error in the else case.

dirkmueller: right, I understand, but do we want to still create the ceph.conf and anything else happening afterwards still in this case? sounds like logging the error and then defaulting to a default pg_num might also be an option. 

Not sure if it really matters though (I'm not torn on this, just wanted to raise the thought as I had it)

tserong: Ah, good thought. So, if the search somehow erroneously came up with no OSDs on one run, but we let it through with a default pg_num, then the config and presumably some services would still be deployed, but with a slightly bogus configuration. Later, at the next chef-client run, if the search fixed itself, the right config would then be written, correct?

mjura: I have also doubts if we really should return here, when pg_num will be 0 ceph.conf can be also generated. During next run when OSDs will be available ceph.conf will be regenerated once again.

mjura:
Looks ok, except `return` in condition

tserong:
OK, I've ditched the return.

tserong:
Can someone trigger another gating run? (Or can I do that somehow?) The last one failed immediately with some pidfile problem :-/

AbelNavarro:
retriggered the build.

@tserong you may be able to retrigger the job yourself if you have a login in ci.suse.de. There's an option in the side menu to do that once you are logged in.

vuntz:
+1. Note that we plan to get this fixed in crowbar later on (because that's really caused by crowbar playing a bit too much with roles).

vuntz:
@tserong needs a rebase after merge of https://github.com/crowbar/crowbar-ceph/pull/49

tserong:
> @tserong needs a rebase after merge of #49

Done. This probably wants backporting to SOC6 too, along with #49. I'm happy to put in both PRs, what's the process for that?

tserong:
Hrm. Gating looks like it's running against crowbar-ceph-4.0+git.1463748899.ddd63f1, so unless I'm mistaken it missed the merge of #49, hence the failure after rebase.

vuntz:
> I'm happy to put in both PRs, what's the process for that?

Just cherry-pick with -x and submit a PR with everything you want :-)

> Hrm. Gating looks like it's running against crowbar-ceph-4.0+git.1463748899.ddd63f1, so unless I'm mistaken it missed the merge of #49, hence the failure after rebase.

Yeah, that's because the package hasn't been updated yet, so gating will fail unless someone manually updates the package (or until tomorrow). I'm updating the packages right now, and I'll re-trigger gating afterwards.

vuntz:
+1

tserong:
OK, how do I find out what codeclimate is doing? It's allgedly "analyzing this code", but if I click the "Details" link, then "Builds", all I can find is commit 60fb3884, which I assume is commit from before the rebase. There doesn't seem to be a mention of 76f1c52.

vuntz:
@tserong just ignore codeclimate ;-) Sometimes it's stalling and we don't know why.

mjura:
+1

PR #49, opened 2016-04-28
Set default pg num based based on number of expected pools (bsc#948522)

tserong:
This sets "osd pool default pg num" based on the number of OSDs and
number of expected pools, using the logic from http://ceph.com/pgcalc/

It's not perfect, but does result in "about 100" PGs per OSD, regardless
of the number of OSDs and pools, so we should no longer get a "suboptimal"
number of PGs per OSD (bsc#948522), nor an "absurd" number of PGs per OSD
(as reported in bsc#968159, a duplicate of the former bug).

It's still not ideal, because if you look at http://ceph.com/pgcalc/
that assumes that some of the RGW pools will use lots of data, and some
will use little data, so it might be best, for RGW deployments, to
pre-create all the RGW pools rather than allowing them to be
automatically created by the radosgw daemon with the default pg num.
OTOH, that would mean making the barclamp aware of the different RGW
pool names depending on which version of Ceph it's deploying (adding
more comlexity), and probably _still_ wouldn't give a perfect result.

TL;DR: the "osd pool default pg num" setting should be markedly better
now than it was before, and RGW deployments won't result in too many PGs
per OSD.  For truly fine control, one should deploy Ceph, then manually
create the RGW poools, then deploy radosgw.

### DIFF
--- a/chef/cookbooks/ceph/recipes/conf.rb
+++ b/chef/cookbooks/ceph/recipes/conf.rb
@@ -64,11 +64,14 @@ end
 expected_pools = 1
 
 rgw_host = search(:node, "roles:ceph-radosgw")
-# According to http://docs.ceph.com/docs/master/radosgw/config/, radosgw uses
-# 13 pools.  http://ceph.com/pgcalc/ suggests radosgw uses 12 pools.  Random
-# experimentation suggests it might actually be 11 pools (at least on Ceph
-# Jewel).  So let's pretend it's 12.
-expected_pools += 12 unless rgw_host.empty?
+# RGW will use up to 14 pools (try "RGW only" with http://ceph.com/pgcalc/)
+# These are not all created immediately though.  Six will be created when
+# the radosgw daemon starts for the first time.  The swift pool(s) will be
+# created when you create a swift subuser for the first time, and the usage
+# pool will only be created if the usage log is explicitly enabled.  Here,
+# to be conservative, we're assuming the full 14 pools will be used for RGW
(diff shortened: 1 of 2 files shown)


chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -59,22 +59,34 @@
...
+# There'll always be at least the default rbd pool (which Ceph always creates
+# with 64 PGs, irrespective of what "osd pool default pg num" is set to)
+expected_pools = 1
+
+rgw_host = search(:node, "roles:ceph-radosgw")
+# According to http://docs.ceph.com/docs/master/radosgw/config/, radosgw uses
+# 13 pools.  http://ceph.com/pgcalc/ suggests radosgw uses 12 pools.  Random
+# experimentation suggests it might actually be 11 pools (at least on Ceph
+# Jewel).  So let's pretend it's 12.
+expected_pools += 12 unless rgw_host.empty?
```
theanalyst: if you select RGW _only_ in pgcalc, you would see 14 pools, 2 of which (.usage, buckets.extra) are not created by default, also most pools are created during run time, for eg the swift pool is created on the fly when you create a swift subuser for the first time, the usage pool is only created if you explicitly configure `rgw enable usage log` in ceph.conf, etc

theanalyst:
lgtm

vuntz:
+1

vuntz:
@tserong will you create the backport PR?

PR #51, opened 2016-05-20
Avoid using the keystone admin token

rhafer:
This requires creating a separate service user for ceph, like we do for all the
other openstack services as well.


rhafer:
This depends on https://github.com/crowbar/crowbar-openstack/pull/366

Also it's still lacking a schema migration for the new service_user and service_password attributes.

chef/data_bags/crowbar/migrate/ceph/100_add_keystone_service_user.rb
```diff
@@ -0,0 +1,25 @@
+# Using a class instance variable to make sure to set the same password
+# in the proposal and in the role
+unless defined?(@ceph_service_password)
+  @ceph_service_password = nil
+end
```
rhafer, line 5: Note: I used this approach instead of a "normal" Class Variable (like e.g. here: https://github.com/crowbar/crowbar-openstack/blob/master/chef/data_bags/crowbar/migrate/rabbitmq/001_generate_password.rb#L5) because the HoundCI was complaining about using Class Variables (https://github.com/crowbar/crowbar-ceph/pull/51#discussion-diff-64227604)

Though I am somehow mixed about which implementation is really cleaner since Class Variables are only realy problematic when inheritance is involved (which is not the case here).

rhafer:
@jgrassler Since this is basically a followup of https://github.com/crowbar/crowbar-openstack/pull/366 I guess it makes sense for you to review it.

jgrassler:
+1

vuntz:
+1

PR #56, opened 2016-06-27
chef: Use new CrowbarRoleRecipe.node_state_valid_for_role helper

vuntz:
This is used in role recipes to decide whether the role should be run or
not. This will allow us to stop rebuilding the run_list based on the
state of the nodes, since we move the logic in the role recipes.


jsuchome:
+1

rsalevsky:
+1

PR #58, opened 2016-07-20
Create base of Cloud 7 Upgrade API V2 for crowbar-ceph

MaximilianMeister:


crowbar_framework/config/routes.d/api_v2_storage.routes
```diff
@@ -0,0 +1,11 @@
+namespace :api,
```
rsalevsky: I would name the file just `storage.routes` because we will have all storage related routes here.

rsalevsky:
+1

Itxaka:
+1

MaximilianMeister:
stripped the `/v2` from the url. the version will be sent now through the accept header

rsalevsky:
+1

PR #62, opened 2016-08-10
Simplify searching for ceph-mon nodes

jsuchome:
I'm deleting some code here which I not fully understand, so it might be really wrong. It is an attempt to find a fix for bug 992392.

But for example that extra search for `ceph_bootstrap-osd-secret:*` seems unnecessary to me, as (according to https://github.com/crowbar/crowbar-ceph/blob/master/chef/cookbooks/ceph/attributes/default.rb#L6) all nodes should already have it


jsuchome:
OK, this does not help :-(

PR #63, opened 2016-08-10
Repeat the search for ceph-mon nodes even after initial failure.

jsuchome:
It seems that the search does not return mon nodes for some initial time.

Ideally, this whole search mechanism should be moved to convergence phase, but that would require larger rewrite of many recipes...


vmoravec:
Tested twice on qa hw with success.
+1

AbelNavarro:
+1

nicolasbock:
+1

PR #65, opened 2016-08-18
Revert "No need to narrow the search for ceph_bootstrap-osd-secret"

rhafer:
This reverts commit 2b3108ed00995061f312ab4a3d93ca8f86c606dc.
It seems that without that additional search criteria the search sometime
returns node with "bootstrap-osd-secret" set to nil.


rhafer:
Example for such a failure: https://ci.suse.de/job/openstack-mkcloud/30964/

(Note, I could not succesfully reproduce that failure. It seems to be extremely racey. But I suggested we add this change and continue watching the CI for future failures)

jsuchome:
+1

vuntz:
+1

PR #66, opened 2016-08-24
Fix port number for radosgw with ssl

nicolasbock:
When using ssl, it is necessary to append a literal 's' to the port number of
the frontend.


tserong:
LGTM

(The original code did work -- it had the "s" inside an `<% if (@rgw_pemfile) %>` guard in the config file, so would only be added if the pemfile was set, which will only be true if ssl is enabled -- but that was a bit obscure, I think what you've done here makes it much clearer)

vuntz:
+1

vuntz:
Counting @tserong's comment as a +1 :-)

PR #67, opened 2016-08-30
Add support for deploying Ceph MDS (fate#321033)

tserong:
This is a rough minimum, still needs some tweaks (see the TODOs),
and definitely needs more testing, but it _does_ work ;)

Open questions:
- Do we require the MDS to run on a separate node, or can it be
  colocated with a MON or OSD?
- Are we allowed to deploy multiple MDSes?
  - If so, how many?  There's no limit enforced presently.
- Should we expose the cephfs filesystem name, and data and metadata
  pool names to the user for editing in the GUI?
- Should we suggest a node for the MDS role by default (as is done
  for rgw?)

Possible problems:
- If you upgrade a system with a ceph proposal that already exists,
  the ceph-mds role isn't made available in the UI, but if you delete
  the proposal then re-create it, the ceph-mds role appears.
  Presumably we'd just want the role to appear as available so it
  could be added to an existing proposal (I shall assume I have missed
  some relevant piece of magic somewhere, or screwed up testing and
  need to re-test).

### DIFF
--- a/chef/cookbooks/ceph/recipes/mds.rb
+++ b/chef/cookbooks/ceph/recipes/mds.rb
@@ -1,0 +1,73 @@
+#
+# Copyright 2016, SUSE LINUX GmbH
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+if node.roles.include?("ceph-mon")
+  include_recipe "ceph::default"
+  include_recipe "ceph::conf"
+else
+  include_recipe "ceph::keyring"
+end
+include_recipe "ceph::server"
+
+directory "/var/lib/ceph/mds/ceph-#{node["hostname"]}" do
+  owner "ceph"
+  group "ceph"
+  mode "0750"
+  recursive true
+  action :create
+end
+
+execute "create mds keyring" do
+  command "ceph auth get-or-create mds.#{node["hostname"]} \
+             osd 'allow rwx' mds 'allow' mon 'allow profile mds' \
+             -o /var/lib/ceph/mds/ceph-#{node["hostname"]}/keyring ; \
+           chown ceph.ceph /var/lib/ceph/mds/ceph-#{node["hostname"]}/keyring"
+  not_if { File.exist?("/var/lib/ceph/mds/ceph-#{node["hostname"]}/keyring") }
+end
+
+service "ceph_mds" do
+  service_name "ceph-mds@#{node["hostname"]}"
+  supports restart: true, status: true
+  action [:enable, :start]
+  subscribes :restart, resources(template: "/etc/ceph/ceph.conf")
+end
+
+service "ceph-mds.target" do
+  action :enable
+end
+service "ceph.target" do
+  action :enable
+end
+
+execute "create data pool" do
+  # Using `rados mkpool` in order to have it pick up the default pg_num from
+  # ceph.conf (compare `ceph osd pool create`, which requires the pg_num to
+  # be specified on the command line)
+  command "rados mkpool #{node[:ceph][:cephfs][:data_pool]}"
+  not_if "rados lspools|grep -q '^#{node[:ceph][:cephfs][:data_pool]}$'"
+end
+
+execute "create metadata pool" do
+  command "rados mkpool #{node[:ceph][:cephfs][:metadata_pool]}"
+  not_if "rados lspools|grep -q '^#{node[:ceph][:cephfs][:metadata_pool]}$'"
+end
+
+execute "create cephfs filesystem" do
+  # NOTE: not_if condition will be fragile if `ceph fs ls` output ever changes
+  command "ceph fs new #{node[:ceph][:cephfs][:fs_name]} \
+             #{node[:ceph][:cephfs][:metadata_pool]} #{node[:ceph][:cephfs][:data_pool]}"
+  not_if "ceph fs ls|grep -q 'name: #{node[:ceph][:cephfs][:fs_name]},"
+end
--- a/chef/data_bags/crowbar/template-ceph.json
+++ b/chef/data_bags/crowbar/template-ceph.json
@@ -40,6 +40,11 @@
           "certfile": "/etc/apache2/ssl.crt/calamari-server.crt",
           "keyfile": "/etc/apache2/ssl.key/calamari-server.key"
         }
+      },
+      "cephfs": {
+        "fs_name": "cephfs",
+        "data_pool": "cephfs_data",
+        "metadata_pool": "cephfs_metadata"
       }
     }
   },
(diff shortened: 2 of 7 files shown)


tserong:
Let's see... Who might like to offer some opinions on this?

From the cloud team, I'm guessing one or more of: @jsuchome, @mjura, @nicolasbock, @dirkmueller

From the storage team, I'm guessing one or more of @l-mb, @jan--f, @smithfarm, @theanalyst

Apologies for the wide circulation, I'm attempting to solicit good answers to the open questions list in the commit message ;-)

smithfarm:
> Are we allowed to deploy multiple MDSes?

No. Multi-MDS is still in development upstream and is not supported.

chef/cookbooks/ceph/recipes/mds.rb
```diff
@@ -0,0 +1,47 @@
...
+service "ceph.target" do
+  action :enable
+end
+
+# TODO: Not sure if the below will be racey or otherwise have timing issues
+# with MDS starting up or pool creation.
+
+execute "create data pool" do
+  # TODO: set sensible pg num
+  command "ceph osd pool create #{node[:ceph][:cephfs][:data_pool]} 256"
```
theanalyst: do we have any defaults for `ceph osd default pg{p} num`; can't we defer to this value instead of 256?

jan--f:
>  Multi-MDS is still in development upstream and is not supported.

If I'm not completely mistaken mutiple MDS's shoudl be possible. Multi MDS is marked experimental in active-active configurations. Active-passive configurations should be no problem and we should probably offer them for failover scenarios.

chef/cookbooks/ceph/recipes/mds.rb
tserong: Yeah, I just threw 256 in there for a test. There's already some code that tries to generate a sensible default in chef/cookbooks/ceph/recipes/conf.rb, and actually saves that to `osd pool default pg{p} num` in /etc/ceph/ceph.conf, so I need to presumably pick that up somehow for this command invocation.

jan--f:
> Do we require the MDS to run on a separate node, or can it be colocated with a MON or OSD?

It's generally recommended to run each daemon on it's own host. On the other hand nothing prevents co-hosting multiple daemons. Imho the OSD/MDS combination should be avoided. This would likely make the FS access very unreliable since metadata and data requests might go to the same machine.
MON/MDS should be ok in some scenarios. Both daemons are memory hungry, so it should realy only be done when lots of RAM is available. CPU is of course also a concern...the MDS is essentially single threaded.

tserong:
> Active-passive configurations should be no problem and we should probably offer them for failover scenarios.

OK, AIUI from a quick read of the docs, enabling multi active MDS requires explicitly increasing the "max_mds" setting and enabling "allow_multimds", so by default if we allow more than one MDS it'll be active-passive, right?

jan--f:
> so by default if we allow more than one MDS it'll be active-passive, right?

That is my understanding too. Even if the passive MDS is not told which rank to handle, the MON cluster will assign it a rank should the active MDS fail.

chef/cookbooks/ceph/recipes/mds.rb
theanalyst: Ah ok, if the default is set I guess just invoking ceph osd pool create should pick that up by itself, iirc

chef/cookbooks/ceph/recipes/mds.rb
```diff
@@ -0,0 +1,47 @@
+if node.roles.include?("ceph-mon")
+  include_recipe "ceph::default"
+  include_recipe "ceph::conf"
+else
+  include_recipe "ceph::keyring"
+end
+include_recipe "ceph::server"
+
+execute "create mds keyring" do
+  command "ceph auth get-or-create mds.#{node["hostname"]} osd 'allow rwx' mds 'allow' mon 'allow profile mds' -o /etc/ceph/mds.#{node["hostname"]}.keyring ; chown ceph.ceph /etc/ceph/mds.#{node["hostname"]}.keyring"
```
jan--f: Should it not say `mds 'allow *` instead of `mds 'allow'`? I see that the ceph doc has an example like this, but in other places the star is explicitly listed.

smithfarm: @jan--f http://tracker.ceph.com/issues/16255 might be relevant? (I don't pretend to understand it)

chef/cookbooks/ceph/recipes/mds.rb
```diff
@@ -0,0 +1,47 @@
+if node.roles.include?("ceph-mon")
```
vuntz, line 17: Copyright header? :-)

vuntz:
> Possible problems:
> -  If you upgrade a system with a ceph proposal that already exists, the ceph-mds role isn't made available in the UI, but if you delete the proposal then re-create it, the ceph-mds role appears. Presumably we'd just want the role to appear as available so it could be added to an existing proposal (I shall assume I have missed some relevant piece of magic somewhere, or screwed up testing and need to re-test).

I think that's because you don't have the schema revision bump in the json file (while you have the migration script ready).

tserong:
> I think that's because you don't have the schema revision bump in the json file (while you have the migration script ready).

Ah! Of course! Thank you. Shall fix and re-test.

chef/cookbooks/ceph/recipes/mds.rb
jan--f: So here is what I think to understand: As mentioned in the issue the `*` permission is needed to issue `mds tell` commands to an MDS. `tell` is used to change config parameters of an MDS while it is running...so the permissions should be ok. I assume an MDS won't change its own config or that of other MDS's.

tserong: > I assume an MDS won't change its own config or that of other MDS's.

OK, I'll leave it as `mds allow` without the `*`.

chef/cookbooks/ceph/recipes/mds.rb
tserong: I've changed it to `rados mkpool`, which picks up the default (`ceph osd pool create` always requires pg_num be specified on the command line)

tserong:
OK, I think I've addressed everything except for a couple of "line too longs". If there's general happiness with this now, I'll squash it back to one commit.

chef/data_bags/crowbar/template-ceph.json
```diff
@@ -40,32 +40,40 @@
...
+        "metadata_pool": "cephfs_metadata"
       }
     }
   },
   "deployment": {
     "ceph": {
       "crowbar-revision": 0,
       "crowbar-applied": false,
-      "schema-revision": 100,
+      "schema-revision": 200,
```
toabctl: 101 is correct I think. For SOC7 we started in all barclamps with 100

tserong:
OK, what do I do when I need a horribly long `command` invocation, but I want Hound to STFU about it? :-)

tserong:
Squashed back to one commit. I'm basically happy with this now.

vuntz:
LGTM, but I know nothing about MDS ;-)

tserong:
Actually, wait, I realised last night that the keyring really belongs under `/var/lib/ceph/mds` for consistency with the other daemons and other deployment methods.

tserong:
OK, I've moved the keyring to where it should have been in the first place, and I was going to say _now_ I'm happy...  But what the hell is a "missing frozen string literal comment"?

tserong:
This also needs a tweak to allow deployment on SP2 (see https://github.com/crowbar/crowbar-ceph/pull/70), so le'ts wait 'til that's merged, then I'll update this to match.

tserong:
Rebased on latest master (Hound violations mercifully gone)

toabctl (approved):

dirkmueller (approved):

PR #68, opened 2016-08-31
enable API endpoint to check the storage repos

MaximilianMeister:
@vuntz @jsuchome as I understood we also need to check for the system repos here (sp2), right?

If the check is correct I would like to move it to `crowbar-core` somewhere, where it can be reused, when we add the checks to `crowbar-ha`.

and the other thing is, if the repos are not enabled/active I guess we just activate them, right? at least that's how we did it in the last upgrade from 5 to 6

depends on: ~~https://github.com/crowbar/crowbar-core/pull/633~~ https://github.com/crowbar/crowbar-core/pull/644

### DIFF
--- a/crowbar_framework/app/controllers/api/storages_controller.rb
+++ b/crowbar_framework/app/controllers/api/storages_controller.rb
@@ -15,6 +15,8 @@
 #
 
 class Api::StoragesController < ApiController
+  before_action :set_ceph
+
   api :GET, "/api/storages", "List all Ceph storages"
   api_version "2.0"
   def index
@@ -31,6 +33,12 @@ class Api::StoragesController < ApiController
   api :GET, "/api/storages/repocheck", "Sanity check ceph repositories"
   api_version "2.0"
   def repocheck
-    render json: {}, status: :not_implemented
+    render json: @ceph.repocheck
+  end
+
+  protected
+
+  def set_ceph
+    @ceph = Api::Storage.new
   end
 end
--- a/crowbar_framework/app/models/api/storage.rb
+++ b/crowbar_framework/app/models/api/storage.rb
@@ -15,6 +15,9 @@
 #
 
 module Api
-  class Storages
+  class Storage < Tableless
+    def repocheck
+      Api::Node.new.repocheck(addon: "ceph")
+    end
   end
 end


MaximilianMeister:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ncdOGOzP&#x2F;97-3-s47p11-implement-api-for-repo-checks-during-upgrade-upgrade-step-8"> (3) S47P11 Implement API for repo checks during upgrade (upgrade step 8)</a>

jsuchome:
Hm, I think the check for system repos should not be part of storage model.

>  and the other thing is, if the repos are not enabled/active I guess we just activate them, right? at 
> least that's how we did it in the last upgrade from 5 to 6

I would say yes.

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,43 @@
...
+    def repocheck
+      {}.tap do |ret|
+        [:ceph, :system].each do |r|
+          ret[r] = {
+            available: false
+          }
+        end
+
+        # check for the system repos
+        ["sles12-sp2-pool", "sles12-sp2-updates"].each do |repo|
```
vuntz: I'm a bit concerned that we're slowly hardcoding repos names everywhere, which will be a pain later on. Can't we somehow use some metadata from the repo yaml file instead to find the repos that need to be there?

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,43 @@
...
+          NodeObject.all.map(&:architecture).each do |arch|
+            repository = Crowbar::Repository.where(
+              platform: "suse-12.2",
```
vuntz: Similarly, hardcoding the platform everywhere... Maybe we should simply have a helper in crowbar-core that gives us the target platform when upgrading nodes?

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,43 @@
...
+              repo: repo
+            )
+            ret[:system][:available] = repository.active?
+          end
+        end
+
+        # check for the storage repo
+        return ret unless Crowbar::Repository.provided?("ceph")
+
+        unless Crowbar::Repository.provided_and_enabled?("ceph")
```
vuntz: Don't we need to check if it's provided for a specific platform?

crowbar_framework/app/models/api/storage.rb
MaximilianMeister: you mean adding metadata to the yaml file, or taking something existing?
btw I moved this code to https://github.com/crowbar/crowbar-core/pull/633
lets continue there, as this is not the right repo

crowbar_framework/app/models/api/storage.rb
MaximilianMeister: okay

crowbar_framework/app/models/api/storage.rb
MaximilianMeister: ah right, we can specify a platform

jsuchome:
I'm still not convinced why would method from ceph API need to call a check for system repositories. I think the check for system/cloud repos needs to be executed independently on the check for ceph repos.
EDIT this is not relevant any more, we've discussed it after

MaximilianMeister:
regarding https://github.com/crowbar/crowbar-core/pull/633#issuecomment-244297653 wouldn't it make sense to somehow determin the necessary architecture here, and pass it over to `provided_and_enabled_with_repolist` ? otherwise we check for repos that we don't need to care about.

crowbar_framework/app/controllers/api/storages_controller.rb
```diff
@@ -15,6 +15,8 @@
 #
 
 class Api::StoragesController < ApiController
+  before_action :set_storage
```
vuntz: (completely out-of-topic, but I think we should have named everything "Ceph", not "Storage", as the repo is called crowbar-ceph)

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,27 @@
...
+
+        ["os", "ceph"].each do |feature|
+          unless ::Crowbar::Repository.provided_and_enabled?(feature, platform)
+            Openstack::Upgrade.enable_repos_for_feature(feature, Rails.logger)
+          end
+          available, repolist = ::Crowbar::Repository.provided_and_enabled_with_repolist(
+            feature, platform
+          )
+          ret[feature]["available"] = available
+          ret[feature]["repos"] = repolist.deep_stringify_keys
```
vuntz: So I would still simplify the return value. To me, `ret` should only include a `ceph` attribute. So that would be something like:

``` ruby
        ret["ceph"] = {}

        platform = Api::Upgrade.new.target_platform(platform_exception: :ceph)

        ["os", "ceph"].each do |feature|
          unless ::Crowbar::Repository.provided_and_enabled?(feature, platform)
            Openstack::Upgrade.enable_repos_for_feature(feature, Rails.logger)
          end
          available, repolist = ::Crowbar::Repository.provided_and_enabled_with_repolist(
            feature, platform
          )
          ret["ceph"]["available"] &&= available
          ret["ceph"]["repos"].deep_merge(repolist.deep_stringify_keys)
        end
```

(the `deep_merge` bit is probably not correct, but hopefully it gives the idea)

The point here is that when we check for ceph, the return value that is expected is only for ceph, and yes, it might be that for ceph, the OS repos are missing.

Does it make sense?

crowbar_framework/app/controllers/api/storages_controller.rb
MaximilianMeister: right, will change it here, for the controller name, we have to see if that creates namespacinng problems, as we already have a ceph_controller. I would prefer to test it in another PR

crowbar_framework/app/models/api/storage.rb
MaximilianMeister: > To me, ret should only include a ceph attribute.

right, that should be enough

> Does it make sense?

it does :-)

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,32 @@
 #
 
 module Api
-  class Storages
+  class Storage < Tableless
+    def repocheck
+      {}.tap do |ret|
+        ret["ceph"] = {}
+        ceph_node_architectures = NodeObject.find(
+          "roles:ceph-mon AND ceph_config_environment:*"
```
MaximilianMeister: @jsuchome @vuntz is that chef search correct to find all related ceph nodes?

jsuchome: I think checking for one node is enough.

MaximilianMeister: okay, so all ceph nodes have the same arch?

jsuchome: Ah, you re looking for arch, not system version. Hm. Actually I have no idea about clouds with possible different architectures...

vuntz: It will very likely not happen for ceph, but for openstack, it's different, and since I assume there'll be some copy & paste, I think it's better to keep the generic code :-)

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,32 @@
...
+                                                               platform,
+                                                               architecture)
+              ::Openstack::Upgrade.enable_repos_for_feature(feature, Rails.logger)
+            end
+            available, repolist = ::Crowbar::Repository.provided_and_enabled_with_repolist(
+              feature, platform, architecture
+            )
+            ret["ceph"]["available"] = available
+            ret["ceph"]["repos"] ||= {}
+            ret["ceph"]["repos"].deep_merge!(repolist.deep_stringify_keys)
```
vuntz: I guess you need to put the architecture somewhere in the hash, to report that to the user.

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,32 @@
...
+          ceph_node_architectures.each do |architecture|
+            unless ::Crowbar::Repository.provided_and_enabled?(feature,
+                                                               platform,
+                                                               architecture)
+              ::Openstack::Upgrade.enable_repos_for_feature(feature, Rails.logger)
+            end
+            available, repolist = ::Crowbar::Repository.provided_and_enabled_with_repolist(
+              feature, platform, architecture
+            )
+            ret["ceph"]["available"] = available
```
vuntz: `=` should probably be `&&=` to not lose the previous value. (it needs to be initialized to `true` somewhere, of course)

crowbar_framework/app/models/api/storage.rb
vuntz: And to answer your initial question: you won't find all ceph nodes this way, just some of them. It'd be better to find all applied ceph proposals, and then look at the elements in there.

crowbar_framework/app/models/api/storage.rb
MaximilianMeister: it is already included in `repolist` see https://github.com/crowbar/crowbar-core/pull/633

vuntz: Ah, ok. I thought we possibly wanted to also have a per-arch `available` status, but it's probably not worth it.

MaximilianMeister:
@vuntz updated

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,44 @@
...
+            ret["ceph"]["repos"].deep_merge!(repolist.deep_stringify_keys)
+          end
+        end
+      end
+    end
+
+    protected
+
+    def ceph_node_architectures
+      ceph_proposal = Proposal.find_by(barclamp: "ceph", name: "default")
```
vuntz: I was careful to say "applied proposal" (I hope ;-)). You're looking at proposals, which are not necessarily applied.

Also, you should not restrict on the name, just in case we change the code to support multiple ceph proposals again (or in case someone named the proposal differently than "default").

MaximilianMeister: ah okay, the "applied" bit got lost somehow, my bad...

MaximilianMeister:
@vuntz updated

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,45 @@
...
+            ret["ceph"]["repos"].deep_merge!(repolist.deep_stringify_keys)
+          end
+        end
+      end
+    end
+
+    protected
+
+    def ceph_node_architectures
+      ceph_proposal = Proposal.find_by(barclamp: "ceph")
```
vuntz: I'd really do a loop on all proposals from `Proposal.where(barclamp: "ceph")`, and not assume there's only one proposal.

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,45 @@
...
+      return [] if ceph_proposal.nil?
+      return [] unless ceph_proposal.status == "ready"
```
vuntz: (btw, interesting trick to find the applied proposals; I would have gone a different approach, by looking at RoleObject, but this may be better)

crowbar_framework/app/models/api/storage.rb
```diff
@@ -15,6 +15,45 @@
...
+
+      ceph_node_names = ceph_proposal.properties["deployment"]["ceph"]["elements"]["ceph-mon"]
```
vuntz: Not just `ceph-mon` :-) Really, take all roles in the ceph proposals.

jsuchome: I must say, I originally suggested to look at just the first ceph-mon node. I did not imagine there could be different architectures inside ceph clusters

crowbar_framework/app/models/api/storage.rb
MaximilianMeister: hmm maybe not, (see https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/proposal.rb#L132) can you paste an example here with the RoleObject?

crowbar_framework/app/models/api/storage.rb
MaximilianMeister: ah right, that's what we discussed earlier

crowbar_framework/app/models/api/storage.rb
vuntz: I guess something like `RoleObject.all.select { |r| r.proposal? }`

MaximilianMeister:
@vuntz updated again

MaximilianMeister:
depends on https://github.com/crowbar/crowbar-core/pull/644 now

rsalevsky:
+1

MaximilianMeister:
@jsuchome the dependent PR is merged now, could you have another look?

jsuchome:
+1

PR #71, opened 2016-09-15
rename storages to ceph

MaximilianMeister:
the barclamp name fits better than the abstracted name

### DIFF
--- a/crowbar_framework/app/controllers/api/ceph_controller.rb
+++ b/crowbar_framework/app/controllers/api/ceph_controller.rb
@@ -14,23 +14,12 @@
 # limitations under the License.
 #
 
-class Api::StoragesController < ApiController
-  api :GET, "/api/storages", "List all Ceph storages"
-  api_version "2.0"
-  def index
-    render json: [], status: :not_implemented
-  end
-
-  api :GET, "/api/storages/:id", "Show a single Ceph storage"
-  param :id, Integer, desc: "Ceph Storage ID", required: true
-  api_version "2.0"
-  def show
-    render json: {}, status: :not_implemented
-  end
-
-  api :GET, "/api/storages/repocheck", "Sanity check ceph repositories"
-  api_version "2.0"
-  def repocheck
-    render json: Api::Storage.repocheck
+module Api
+  class CephController < ApiController
+    api :GET, "/api/ceph/repocheck", "Sanity check ceph repositories"
+    api_version "2.0"
+    def repocheck
+      render json: Api::Ceph.repocheck
+    end
   end
 end
--- a/crowbar_framework/app/models/api/ceph.rb
+++ b/crowbar_framework/app/models/api/ceph.rb
@@ -15,7 +15,7 @@
 #
 
 module Api
-  class Storage < Tableless
+  class Ceph < Tableless
     class << self
       def repocheck
         Api::Node.repocheck(addon: "ceph")
--- a/crowbar_framework/config/routes.d/ceph.routes
+++ b/crowbar_framework/config/routes.d/ceph.routes
@@ -1,7 +1,6 @@
 namespace :api,
   constraints: ApiConstraint.new(2.0) do
-  resources :storages,
-    only: [:index, :show] do
+  resource :ceph, controller: :ceph, only: [] do
     collection do
       get :repocheck
     end


crowbar_framework/app/controllers/api/ceph_controller.rb
```diff
@@ -14,23 +14,22 @@
 # limitations under the License.
 #
 
-class Api::StoragesController < ApiController
+class Api::CephController < ApiController
   before_action :set_ceph
 
-  api :GET, "/api/storages", "List all Ceph storages"
+  api :GET, "/api/ceph", "Show the Ceph object"
```
jsuchome: Hm, I do not like the "object", it looks like something inside ceph storage. What about "configuration"? (Not much better...)

crowbar_framework/app/controllers/api/ceph_controller.rb
```diff
@@ -14,23 +14,22 @@
 # limitations under the License.
 #
 
-class Api::StoragesController < ApiController
+class Api::CephController < ApiController
```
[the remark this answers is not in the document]
aspiers: I think I agree with Hound here. I've seen problems in the past caused by not following this guideline.

crowbar_framework/app/controllers/api/ceph_controller.rb
aspiers: I agree that something less vague would be better. What does the API call return, exactly? Ideally the description should clearly reflect that.

aspiers (changes requested):
Code LGTM (I think - my Rails routes knowledge is a bit out of date) - just one or two minor nitpicks.

crowbar_framework/app/controllers/api/ceph_controller.rb
MaximilianMeister: right now nothing, but it should display some general data (see also GET /api/crowbar). what about returning the ceph version for a start?

crowbar_framework/app/controllers/api/ceph_controller.rb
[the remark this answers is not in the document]
MaximilianMeister: added to cleanup card https://trello.com/c/8p1DvJQ2/1931-cleanup-upgrade-related-api as we have to change it in multiple places (in other files/repos as well)

crowbar_framework/app/controllers/api/ceph_controller.rb
MaximilianMeister: I just saw that we don't store the ceph version anywhere as far as i could see. but lets just drop this endpoint for now as we don't really have a use case

MaximilianMeister:
@aspiers @jsuchome updated (not sure though why the comments are duplicated, and the hound doesn't disappear)

jsuchome:
+1

dirkmueller (approved):

aspiers:
@MaximilianMeister It seems that github's new features still have some bugs to iron out ;-)

crowbar_framework/app/models/api/ceph.rb
```diff
@@ -15,7 +15,7 @@
 #
 
 module Api
-  class Storage < Tableless
+  class Ceph < Tableless
     def repocheck
```
aspiers: Why is this an instance method rather than a class method?

crowbar_framework/app/controllers/api/ceph_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+    api :GET, "/api/ceph/repocheck", "Sanity check ceph repositories"
+    api_version "2.0"
+    def repocheck
+      render json: @ceph.repocheck
+    end
+
+    protected
+
+    def set_ceph
+      @ceph = Api::Ceph.new
```
aspiers: I know this PR isn't introducing or even really changing this, but I don't understand why it's an instance method or why an instance variable is being assigned, since I can't see any instance-specific state.  Am I missing something?

aspiers (commented):
The rename looks fine but I'm a bit puzzled about the design of the code which already existed before this PR.

crowbar_framework/app/controllers/api/ceph_controller.rb
MaximilianMeister: good that you mention that, I already thought about this as well. I added a task to the [cleanup trellocard](https://trello.com/c/8X7elkka/82-cleanup-upgrade-related-mess)

MaximilianMeister: @aspiers see https://github.com/crowbar/crowbar-ceph/pull/72

MaximilianMeister:
rebased and added an example to the api doc (according to https://github.com/crowbar/crowbar-core/pull/682)

jsuchome (approved):
+1

rsalevsky (changes requested):
Can you splitt this change into two separate PRs? This are two totally different topics.

MaximilianMeister:
> Can you splitt this change into two separate PRs

dropped the commit

rsalevsky (approved):

MaximilianMeister:
@aspiers ping

aspiers (approved):

PR #73, opened 2016-09-23
Add ceph_client network (fate#319122)

tserong:
This adds a separate ceph client network, instead of having ceph
client traffic go over the admin network.

I'm not clear on exactly what this will do if applied to an
existing cluster that's already using the admin network for client
traffic.  I mean, it'll rewrite ceph.conf to use the new network,
but I haven't tried it agaist an existing cluster to see if it
actually works.

### DIFF
--- a/chef/cookbooks/ceph/libraries/default.rb
+++ b/chef/cookbooks/ceph/libraries/default.rb
@@ -93,7 +93,11 @@ def get_mon_addresses()
 
     mons += get_mon_nodes()
     if is_crowbar?
-      mon_ips = mons.map { |node| Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address }
+      mon_ips = mons.map do |node|
+        Chef::Recipe::Barclamp::Inventory.get_network_by_type(
+          node, node["ceph"]["client_network"]
+        ).address
+      end
     else
       if node["ceph"]["config"] && node["ceph"]["config"]["public-network"]
         mon_ips = mons.map { |nodeish| find_node_ip_in_network(node["ceph"]["config"]["public-network"], nodeish) }
@@ -158,8 +162,12 @@ def get_osd_nodes()
     cluster_addr = ""
     public_addr = ""
 
-    public_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
-    cluster_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "storage").address
+    public_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(
+      node, node["ceph"]["client_network"]
+    ).address
+    cluster_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(
+      node, "storage"
+    ).address
 
     osd = {}
     osd[:hostname] = node.name.split(".")[0]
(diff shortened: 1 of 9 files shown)


tserong:
Note that this requires https://github.com/crowbar/crowbar-core/pull/676 be merged first.

tserong:
Ping @mjura, @dirkmueller

tserong:
Doesn't seem to work after an upgrade. Even if I manually create the ceph_client network by editing the network proposal in raw mode, then reapply the ceph barclamp, I get "Failed to apply the proposal: exception before calling chef (undefined method `[]' for nil:NilClass)" because it can't find the ceph_client network for some reason.

Works fine for a new install though...

chef/cookbooks/ceph/libraries/default.rb
```diff
@@ -93,7 +93,9 @@ def get_mon_addresses()
 
     mons += get_mon_nodes()
     if is_crowbar?
-      mon_ips = mons.map { |node| Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address }
+      mon_ips = mons.map do |node|
+        Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "ceph_client").address
```
rhafer: I am not entirely sure about the purpose  of that network, but wouldn't it make sense to be able to reference any of the networks that are defined in the network barclamp here. By make the name an attribute on the ceph proposal? Instead of just hardcoding to "ceph_client"?

That way the user to choose whatever fits their enviroment best.

tserong: Good thought. And, actually, making this a bit smarter with a fallback to "admin" might fix the upgrade problem.

The purpose of the separate network is to not have ceph client traffic go over the admin network, because the assumption is that the admin network will usually be crap (1GbE). Additionally, for storage-only deploys, ceph clients probably don't _want_ to be on the crowbar admin network with its DHCP server.

chef/cookbooks/ceph/libraries/default.rb
```diff
@@ -158,7 +160,7 @@ def get_osd_nodes()
     cluster_addr = ""
     public_addr = ""
 
-    public_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "admin").address
+    public_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(node, "ceph_client").address
```
dirkmueller: This will likely break upgrade.. we need to think about what to do for Cloud6-> upgrade users (unless $bigwig decides that this is not important)

dirkmueller (dismissed):
Let's make the ceph public network configurable (default to admin for the upgrade case, maybe default to public for new deploys and/or when cephfs is required?)

tserong:
Modulo a couple of Hound violations, c8d32aa should be about right. The network to use for ceph client traffic is now an attribute, and defaults to "public" by default for new deployments. This attribute will be set to "admin" during the migration, so there will be no change of behaviour for upgrades (i.e. upgraded deployments keep the client traffic on the admin network). The one kink is that this means you won't be able to use CephFS with Manila on an upgraded system, because the client traffic will still be on the admin network, which is inaccessible to deployed instances. I do not think this is easily fixable though.

A documentation update will be necessary to instruct users that if they don't want the ceph client traffic on the public network, they should edit the ceph proposal in raw mode and set client_network to something other than "public" _before_ deploying ceph (changing it later will break things).

Note: this patch introduces schema revision 102, so applies on top of #67, which introduced schema revision 101.

tserong:
Ha! Take that, Hound!

dirkmueller (approved):

MaximilianMeister (approved):
lgtm

PR #75, opened 2016-11-08
Improve "ceph auth get-or-create" error handling

toabctl:
Return after the first command if that fails to make the error
message visible to chef.


jsuchome:
Is it related to some known bug?

tserong (approved):
LGTM

jsuchome (approved):

nicolasbock (approved):
Nice catch!

dirkmueller (approved):

PR #76, opened 2016-11-10
Allow custom config section in ceph.conf

toabctl:
Use the default["ceph"]["config_sections"] attribute to allow
adding extra sections in ceph.conf .
This is needed eg. for Manila (which uses the ceph.conf)  where the
client used by the manila-share service needs some extra parameters like
"client mount uid" and "client mount gid = 0".

Also remove the unused default["ceph"]["config-sections"] attribute.

### DIFF
--- a/chef/cookbooks/ceph/templates/default/ceph.conf.erb
+++ b/chef/cookbooks/ceph/templates/default/ceph.conf.erb
@@ -141,3 +141,10 @@
   <% end -%>
 <% end -%>
 <% end -%>
+
+<% if (! node['ceph']['config_sections'].nil?) -%>
+  <% node['ceph']['config_sections'].each do |section_name, section_content| -%>
+[<%= section_name -%>]
+<%= section_content %>
+  <% end -%>
+<% end -%>
--- a/chef/cookbooks/ceph/attributes/conf.rb
+++ b/chef/cookbooks/ceph/attributes/conf.rb
@@ -1,5 +1,8 @@
 default["ceph"]["config"] = {}
-default["ceph"]["config-sections"] = {}
 default["ceph"]["config"]["fsid"] = "11dd315a-2cab-4130-a760-b285324ef622"
 default["ceph"]["config"]["osds_in_total"] = 0
 default["ceph"]["config"]["replicas_number"] = 0
+
+# possibility to add extra config sections to ceph.conf
+# where "key" is the section name and value the section content
+default["ceph"]["config_sections"] = {}


chef/cookbooks/ceph/templates/default/ceph.conf.erb
```diff
@@ -141,3 +141,10 @@
   <% end -%>
 <% end -%>
 <% end -%>
+
+<% if (! node['ceph']['config_sections'].nil?) -%>
```
aplanas, line 145: This the value is not nil but a empty hash, so I guess that this will be always `true`?

Maybe `if !node['ceph']['config_sections'].empty?` is more correct

chef/cookbooks/ceph/templates/default/ceph.conf.erb
```diff
@@ -141,3 +141,10 @@
...
+  <% node['ceph']['config_sections'].each do |section_name, section_content| -%>
+[<%= section_name -%>]
+<%= section_content %>
+  <% end -%>
+<% end -%>
```
aplanas: extra line

aplanas (changes requested):

toabctl:
> This the value is not nil but a empty hash, so I guess that this will be always true?

it's not:
```
2.3.0 :001 > {}.nil?
 => false 
2.3.0 :002 > 
```

chef/cookbooks/ceph/templates/default/ceph.conf.erb
toabctl: it's not:
```
2.3.0 :001 > {}.nil?
 => false 
2.3.0 :002 > 
```

toabctl: It catches this case:

```
2.3.0 :003 > x=Hash.new
 => {} 
2.3.0 :004 > x
 => {} 
2.3.0 :005 > x["dd"].nil?
 => true 
```

when no config_sections key is available

chef/cookbooks/ceph/templates/default/ceph.conf.erb
toabctl: done

chef/cookbooks/ceph/templates/default/ceph.conf.erb
aplanas: but you set to `{}` in `chef/cookbooks/ceph/attributes/conf.rb`

aplanas: after talking this @toabctl, we agree the this `.nil?` check is a safety net in case of a random `node['ceph'].delete("config_sections")`

aplanas (approved):

nicolasbock (approved):

dirkmueller (approved):

mjura (commented):
LGTM

PR #77, opened 2016-11-21
Reorder ceph-mds / cephfs setup

dirkmueller:
It seems in many cases ceph-mds is hanging with #cpu's*100% cpu usage.
Trying to order things differently which hopefully avoids the issue


dirkmueller:
It appears this change helps indeed

toabctl (approved):

tserong (approved):
LGTM, although I'd love to know why it pegs the CPU (I didn't see that in my testing), and how moving service start to *after* pool creation helps :-/ @jan--f, have you seen anything like this before?

jan--f:
@tserong Haven't seen this. I don't know how chef does things but I could imagine that the explicit dependency `notifies :run, "execute[create-cephfs]", :immediately` in service ceph-mds eliminates a potential race.

PR #80, opened 2016-12-19
Decrease amount of node saves

vuntz:
Saving is not cheap, and most of the time, it's not needed (as no attribute changed, or because we will already save later).

(Here's it's not critical as the saves only occur on initial setup)

### DIFF
--- a/chef/cookbooks/ceph/recipes/osd.rb
+++ b/chef/cookbooks/ceph/recipes/osd.rb
@@ -69,13 +69,19 @@ else
   end
 
   if is_crowbar?
-    node.set["ceph"]["osd_devices"] = [] if node["ceph"]["osd_devices"].nil?
+    dirty = false
+
+    node.set["ceph"]["osd_devices"] ||= []
     min_size_blocks = node["ceph"]["osd"]["min_size_gb"] * 1024 * 1024 * 2
     unclaimed_disks = BarclampLibrary::Barclamp::Inventory::Disk.unclaimed(node).sort.select { |d| d.size >= min_size_blocks }
 
     # if devices for journal are explicitely listed, do not use automatic journal assigning to SSD
     if !node["ceph"]["osd"]["journal_devices"].empty?
-      node.set["ceph"]["osd"]["use_ssd_for_journal"]        = false
+      # explicit comparison because we don't want a condition that uses nil
+      if node["ceph"]["osd"]["use_ssd_for_journal"] != false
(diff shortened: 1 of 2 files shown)


chef/cookbooks/ceph/recipes/osd.rb
```diff
@@ -69,13 +69,19 @@
   end
 
   if is_crowbar?
-    node.set["ceph"]["osd_devices"] = [] if node["ceph"]["osd_devices"].nil?
+    dirty = false
```
rsalevsky, line 72: I would call it `changed` ;)

matelakat (commented):
@vuntz I see a pattern evolving here (and looking at the other changes as well) Can we make some sort of accessor/proxy/decorator that wraps the node, set attributes on the decorator, and it knows only to save the decorated object if any value changed?

Itxaka:
Agreed with @matelakat here, I was thinking the same thing. Maybe cache the current attributes on the Nodeobject and override the save node to check if the current attributes have changed compared to the cached ones and call the parent save in that case?

Not a ruby expert so this may not be possible or may incur a performance hit to have a cache of attributes?

vuntz:
@matelakat @Itxaka let me think about it over lunch and see if I can come with something.

vuntz:
So I'm torn. I can find ways to do something like suggested, but nothing that will end up looking like:

```ruby
node.set[:a][:b][:c] = foobar
node.save
```

The most simple approach (from the caller perspective) would be something like:

```ruby
node.set_value(:a, :b, :c, foobar)
node.save_if_changed
```

But I kind of dislike the fact that we move away from what looks like standard chef code.

I don't see a good way to simply override `set` in a compatible way (and I know we don't want to override `save` as sometimes we want to really just save as usual, so we may as well get a new method). Of course, I'm not really the best ruby experts out there :-)

Anyone having feelings on this?

aspiers:
It's evil, but it might be possible to monkey-patch `Chef::Node#set` (which is [an alias for `#normal`](https://github.com/chef/chef/blob/10.32.2/chef/lib/chef/node.rb#L263)) so that it always checks to see whether the value being set is different to what it was before, and if so, sets a `dirty` attribute on the node object.  Then it would be trivial to implement `Chef::Node#save_if_dirty` via monkey-patching.

aspiers:
And actually, if we did it right, probably we could just monkey-patch `#save` too, since presumably there's no good reason to save a *pre-existing* node if there are no changes.

PR #81, opened 2016-12-19
Calamari also conflicts with database (bsc#1001946)

dirkmueller:
Calamari messes by itself with the postgresql database, so
sharing that with the rest of the openstack database is
dangerous.


tserong (commented):
LGTM

toabctl (approved):

jsuchome:
@dirkmueller 
I've just found out that the constraint check that uses "conflicts_with" is using only the roles in current barclamp, so this will not work.

PR #82, opened 2017-01-09
ceph: remove certificate conversion

sidthe:
PKI support for keystone has been deprecated in Newton and Ceph shall not rely on keystone token signing capabilities anymore

### DIFF
--- a/crowbar_framework/app/models/ceph_service.rb
+++ b/crowbar_framework/app/models/ceph_service.rb
@@ -52,7 +52,8 @@ class CephService < PacemakerServiceObject
             "suse" => "/^12.*/",
             "opensuse" => "/.*/"
           },
-          "conflicts_with" => ["ceph-mon", "ceph-osd", "ceph-radosgw", "horizon-server"]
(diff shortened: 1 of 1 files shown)


sidthe:
[(3) Remove support for PKI tokens](https://trello.com/c/RJGh4Jg1/394-3-remove-support-for-pki-tokens)

sidthe:
manually tested jointly with the keystone pki related patches applied, as far as I see does the job

vuntz:
I'm not sure this is really correct: my understanding is that we still need to configure this; it just turns out this was done automatically thanks to the code removed here and in https://github.com/crowbar/crowbar-openstack/pull/702 (which was run not just when PKI is used).

@dirkmueller disabled the radosgw tests for the time being in the gating, so failures there are not visible. You may want to revert https://github.com/SUSE-Cloud/automation/pull/1563 locally and see if things still work.

sidthe:
@vuntz Thank you. I accidentally messed up with the commit history here and closed it. Will move to a different PR :(

PR #84, opened 2017-01-16
ceph: try to use the public iface first for ceph-mon service

Itxaka:


Itxaka:
[(5) S53P6 bsc#1018869 - Failed to apply Ceph proposal](https://trello.com/c/t1diZEAO/429-5-s53p6-bsc-1018869-failed-to-apply-ceph-proposal)

chef/cookbooks/ceph/recipes/mon.rb
```diff
@@ -19,8 +19,9 @@
 include_recipe "ceph::conf"
 
 service_type = node["ceph"]["mon"]["init_style"]
+address = node["crowbar"]["network"]["public"] || node["hostname"]
```
vuntz: A couple of things are wrong here :-)

- I don't think you want to hardcode the public network, but reference one defined in the client_network attribute (https://github.com/crowbar/crowbar-ceph/blob/master/chef/data_bags/crowbar/template-ceph.json#L19).

- Do not use `node["crowbar"]["network"]["public"]`, but instead `Barclamp::Inventory.get_network_by_type(node, "public").address`. This is important as the location of the IP address could change.

vuntz (dismissed):

chef/cookbooks/ceph/recipes/mon.rb
Itxaka: This is what happens when you eyeball it while waiting for your cloud to be setup :D

Thanks for the tips @vuntz will redo once I can test the whole thing!

Itxaka:
Still not woking. Again seems to be a problem of the hostname initial member:

```
        ; specify the initial monitors of the cluster in to establish a quorum
        mon initial members = d52-54-77-77-77-01 d52-54-77-77-77-03 d52-54-77-77-77-02
        mon host = 192.168.234.4, 192.168.234.2, 192.168.234.3

```

Those hostnames resolution point to different ips, so there is some kind of issue there, which I dont really understand (Maybe trying to connect to those as initial members before using the `mon host` ones?)

Good thing is, removing the `mon initial members` from all 3 nodes and restarting the ceph-mon, creates a beautiful quorum, so Im not sure how required is that value and if we can just remove it or it has consecuences.

Going by http://docs.ceph.com/docs/master/rados/configuration/mon-config-ref/#initial-members it speeds up the initial quorum, but its not required.

Reading up on http://docs.ceph.com/docs/master/rados/configuration/mon-config-ref/#minimum-configuration it seems that we could get away by dropping those `mon initial members` and using just `mon host` (Althougth we are using them wrong, they should be hostnames, not ips). 

@vuntz @dirkmueller @tserong comments on how to proceed?

vuntz:
@Itxaka change https://github.com/Itxaka/crowbar-ceph/blob/06874ab1b58b3ee87a6cded861e60bf547bfeb83/chef/cookbooks/ceph/recipes/conf.rb#L7 to not use the hostnames, but the hostname on the right network. For instance, for public, this would be public.$hostname.

Itxaka:
@vuntz Im not sure how that works, on my test environment there is only a full hostname (machine.crowbar.g10.cloud.suse) and the short one.

Maybe you meant $hostname.$public ?

dirkmueller:
@Itxaka hmm, there should be a public.* but you're right, there is no hostname for other networks. can you just use the ip address? that would work always

Itxaka:
@dirkmueller done, tested the patch locally and it seems to work.

Itxaka:
while this is working now, it seems that the ceph tests on mkcloud need to be adjusted to use ips instead of hostnames. Working on that

vuntz:
> @Itxaka hmm, there should be a public.* but you're right, there is no hostname for other networks. can you just use the ip address? that would work always

I don't understand: there's always a DNS name for each network a node is on. So if d52-54-77-77-77-01.g9.cloud.suse.de is on the storage network, there'll be a storage.d52-54-77-77-77-01.g9.cloud.suse.de DNS name for this IP address.

So with this in mind... why are we switching to IP addresses? (I might be missing something)

dirkmueller:
@vuntz ah, that also applies for any user provided crowbar network, like "foobarnet" ? ok then, I prefer DNS over IP.

dirkmueller:
@Itxaka could you take a look? this is blocking M8 ..

Itxaka:
thanks @mjura, will test it!

mjura (approved):
Looks and works good

mjura:
It was manually tested and works fine, let's wait for gating tests.

@vuntz: do you still have any comments on this change?

chef/cookbooks/ceph/recipes/conf.rb
```diff
@@ -6,7 +6,14 @@
 
 mon_init = []
 mon_nodes.each do |monitor|
-    mon_init << monitor.name.split(".")[0]
+  mon_name = monitor.name.split(".")[0]
```
dirkmueller: why not using get_node_name here?

dirkmueller (changes requested):

chef/cookbooks/ceph/recipes/conf.rb
mjura: Here we are creating node_name for remote monitors in the cluster, function get_node_name operates on local node object. Call here `monitor.get_node_name` will throw 
```
ArgumentError
-------------
Attribute node is not defined!
```

mjura: I would prefer to keep it, as it is

mjura:
Recheck for gating job, I've checked it locally on my dev environment and tests are passing

mjura:
To make gating test green for this change we have to merge first https://github.com/SUSE-Cloud/automation/pull/1610

Itxaka:
Althougth the automation change has been merged 3 dasy ago, jenkis is still not using it:

```
13:18:24 +(qa_crowbarsetup.sh:4016) ceph_testsuite_configure_storage3(): ./update_template d52-54-77-77-77-01.vl1.cloud.suse.de d52-54-77-77-77-03.vl1.cloud.suse.de d52-54-77-77-77-02.vl1.cloud.suse.de
```

Not sure how long it takes for jenkis to use the latest automation repo?

aplanas:
Restarted the job with ceph:

https://ci.suse.de/job/openstack-mkcloud/50118/console

Itxaka:
@dirkmueller Green like soylent, mind re-reviewing or removing the "change requested"?

Also cc @aplanas :)

aplanas (approved):
This will unlock the gating

chef/cookbooks/ceph/recipes/conf.rb
dirkmueller: sure, but you could just pass in the node as a parameter, and then it would work, e.g. 

```ruby
    get_node_name(monitor)
```

chef/cookbooks/ceph/providers/client.rb
```diff
@@ -28,14 +28,16 @@ def whyrun_supported?
...
   @current_resource.ceph_conf(@new_resource.ceph_conf)
   @current_resource.admin_keyring(@new_resource.admin_keyring)
   @current_resource.name(@new_resource.name)
   @current_resource.as_keyring(@new_resource.as_keyring)
-  @current_resource.keyname(@new_resource.keyname || "client.#{current_resource.name}.#{node['hostname']}")
+  @current_resource.keyname(@new_resource.keyname || "client.#{current_resource.name}.#{node_name}")
   @current_resource.caps(get_caps(@current_resource.ceph_conf, @current_resource.admin_keyring, @current_resource.keyname))
-  default_filename = "/etc/ceph/ceph.client.#{@new_resource.name}.#{node['hostname']}.#{@new_resource.as_keyring ? 'keyring' : 'secret'}"
+  default_filename = "/etc/ceph/ceph.client.#{@new_resource.name}.#{node_name}." \
+                                            "#{@new_resource.as_keyring ? 'keyring' : 'secret'}"
```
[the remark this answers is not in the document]
dirkmueller: I'd like to ignore that one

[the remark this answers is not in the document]
Itxaka: +1

chef/cookbooks/ceph/recipes/conf.rb
dirkmueller: wasn't that difficult to do, so I just pushed that

dirkmueller (approved):

tserong (approved):
LGTM

Itxaka:
Broke it somehow

chef/cookbooks/ceph/recipes/conf.rb
mjura: Sorry, I had to revert this part, because `get_node_name(monitor)` doesn't work here. You look on gating test here ttps://ci.suse.de/job/openstack-mkcloud/50245/consoleFull. Right now gating tests should pass

mjura (approved):

chef/cookbooks/ceph/providers/client.rb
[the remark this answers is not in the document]
mjura: I tried to change it, but it failed
+1 for current state

vmoravec:
Gating passed on ceph, only tempest failed at `TestVolumeBootPatternV2`

PR #86, opened 2017-01-27
ceph: Improve node assignment validation (bsc#1022259)

dirkmueller:
When ceph-mds conflifts with osd, it should reject it independent
of which one is assigned first.


dirkmueller:
we should really fix that in the validation function instead that it is checking both directions, but I don't have the mood to do that now..

Itxaka (approved):

nicolasbock (approved):

rsalevsky (approved):

vuntz (approved):

PR #87, opened 2017-02-01
ceph: Specify the public address for the ceph-mon daemon

Itxaka:
In case there is several different IP addresses in the nodes
under the same subnet, the mon daemon will bind to the first one
and we may not want that. So instead, set specifically the
address that the daemon needs to bind to.

### DIFF
--- a/chef/cookbooks/ceph/recipes/conf.rb
+++ b/chef/cookbooks/ceph/recipes/conf.rb
@@ -3,6 +3,10 @@ raise "fsid must be set in config" if node["ceph"]["config"]["fsid"].nil?
 mon_nodes = get_mon_nodes
 osd_nodes = get_osd_nodes
 mon_addr = get_mon_addresses
+public_addr = Chef::Recipe::Barclamp::Inventory.get_network_by_type(
+  node,
+  node["ceph"]["client_network"]
+).address
 
 mon_init = []
 mon_nodes.each do |monitor|
@@ -114,6 +118,7 @@ template "/etc/ceph/ceph.conf" do
     pool_size: rep_num,
     pool_pg_num: pg_num,
     osd_nodes_count: osd_nodes.length,
+    public_addr: public_addr,
     public_network: node["ceph"]["config"]["public-network"],
     cluster_network: node["ceph"]["config"]["cluster-network"],
     is_rgw: is_rgw,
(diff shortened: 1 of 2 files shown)


Itxaka:
[(5) S53P6 bsc#1018869 - Failed to apply Ceph proposal](https://trello.com/c/t1diZEAO/429-5-s53p6-bsc-1018869-failed-to-apply-ceph-proposal)

dirkmueller:
Ugh, there is a public net parameter? Why didn't we use that from the beginning?!

That means we can revert alle the previous pull request that moved everything to public net?

mjura (approved):
LGTM

Itxaka:
@dirkmueller maybe... some of those were to fix the hostname resolution for the other mons ips so they may still be necessary.

mjura:
I agree with @Itxaka, previous change is also needed, because of proper hostname configuration

Itxaka:
I messed up something as the address was filled empty :(

Itxaka:
duuuuh, was missing a `=`, sorry about that (︶︹︺)

nicolasbock (approved):

Itxaka:
Tested on: https://ci.suse.de/view/Cloud/view/QA/job/cloud-mkphyscloud-qa-scenario-2c/324/console and it went over the ceph part correctly :dancer:

dirkmueller (approved):

dirkmueller:
@Itxaka PR gating is failing, any idea?

tserong:
> In case there is several different IP addresses in the nodes under the same subnet, the mon daemon will bind to the first one and we may not want that.

If there are several IP addresses, and they're all on the same subnet, why does it matter if it binds to the first one? I'd've thought all IP addresses would be equally accessible, if they're on the same subnet? Or is does this cause a problem with hostname resolution? Or something else that I'm missing? :-)

Is this also a problem for OSDs and MDSes or is it only an issue for MONs?

Itxaka:
> If there are several IP addresses, and they're all on the same subnet, why does it matter if it binds to the first one? I'd've thought all IP addresses would be equally accessible, if they're on the same subnet? Or is does this cause a problem with hostname resolution? Or something else that I'm missing? :-)

Should not matter it seems, *but* we are getting the other monitors IP addresses based on the client_network (`node[:ceph][:client_network]`) so ceph ends up listening on one IP but trying to connect to the other monitors in a *different* IP due to the different configuration of both configs.`This works on our mkcloud setups due to how the network is configured but could cause issues on other deployments, as shown on the QA jobs where it totally failed.

This only makes sure that it listens on the expected IP associated with the configured network.

> Is this also a problem for OSDs and MDSes or is it only an issue for MONs?

May affect all of them, but I havent seen the issue on either of those daemons `¯\_(ツ)_/¯`

> @Itxaka PR gating is failing, any idea?

no. retriggered and will follow up :(

tserong (approved):
Thanks for the explanation @Itxaka. I think I'm happy with this :-) If it turns out to be an issue for OSDs or other daemons later, we can deal with it then in the same way, but my guess is, in this context, it's probably only a MON issue.

PR #89, opened 2017-03-03
Fix inconsistent default pg_num when deploying rgw/mds (bsc#1023801)

tserong:
Somehow, searching for nodes with the ceph-radogw and ceph-mds role
doesn't succeed when conf.rb is run the first time on MON and OSD
nodes, which causes the number of expected pools to not be set
correctly on those nodes, which screws up the default pg num in
ceph.conf, resulting in too many PGs per OSD.  Changing it to search
for roles, rather than nodes-with-roles fixes this.

Signed-off-by: Tim Serong <tserong@suse.com>

### DIFF
--- a/chef/cookbooks/ceph/recipes/conf.rb
+++ b/chef/cookbooks/ceph/recipes/conf.rb
@@ -81,7 +81,7 @@ else
   # with 64 PGs, irrespective of what "osd pool default pg num" is set to)
   expected_pools = 1
 
-  rgw_host = search(:node, "roles:ceph-radosgw")
+  rgw_roles = search(:role, "name:crowbar-* AND run_list_map:ceph-radosgw")
   # RGW will use up to 14 pools (try "RGW only" with http://ceph.com/pgcalc/)
   # These are not all created immediately though.  Six will be created when
   # the radosgw daemon starts for the first time.  The swift pool(s) will be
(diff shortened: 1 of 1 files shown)


jsuchome:
Isn't the commit message bit misleading? It does not exactly change `pg_num`, it tries to fix the search, right? (Which I assume will result in correct pg_num later in the code...)

vuntz:
That likely happens because the nodes only get the roles when chef has run.

Better than doing search, you can actually look at the attributes that match the roles assignments. Something like `node[:ceph][:elements]["ceph-radosgw"]`. The result should be the same (assuming the node this is running on is part of the ceph proposal).

tserong:
> Isn't the commit message bit misleading? It does not exactly change pg_num, it tries to fix the search, right? (Which I assume will result in correct pg_num later in the code...)

Yeah, it fixes the search. I was calling it inconsistent pg_num because the broken search means that the MON and OSD nodes will have (for example) "osd pool default pg num = 256" in ceph.conf whereas the MDS and RGW nodes will have "osd pool default pg num = 16", at least initially (later when chef-client has run on all the nodes again they all end up with the same value, but by then it's too late, as pool have been created with the wrong default.

Anyway, I can change the commit message :-)

tserong:
> The result should be the same (assuming the node this is running on is part of the ceph proposal).

This'll end up running on cinder, glance and nova nodes if they're configured to use ceph, which could give ceph.conf on those nodes a "osd pool default pg num" inconsistent with that on the ceph nodes. Practically that probably doesn't actually matter though.

tserong:
Update commit message.

jsuchome (approved):

vuntz (approved):

PR #94, opened 2017-06-02
Do not deploy ceph roles to controllers.

jsuchome:
Mixing ceph roles with openstack ones is not supported scenario,
so let'stry to avoid it at least as automatic placement.

### DIFF
--- a/crowbar_framework/app/models/ceph_service.rb
+++ b/crowbar_framework/app/models/ceph_service.rb
@@ -122,29 +122,41 @@ class CephService < PacemakerServiceObject
     if osd_nodes.size < 2
       osd_nodes_controller = select_nodes_for_role(nodes, "ceph-osd", "controller")
       osd_nodes_all = select_nodes_for_role(nodes, "ceph-osd")
-      # avoid controllers if possible (will likely be used for mon)
+      # avoid controllers if possible (ceph should not be used with openstack roles)
       osd_nodes_no_controller = osd_nodes_all - osd_nodes_controller
-      osd_nodes = [osd_nodes, osd_nodes_no_controller, osd_nodes_all].flatten.uniq{ |n| n.name }
+      osd_nodes = [osd_nodes, osd_nodes_no_controller, osd_nodes_all].flatten.uniq(&:name)
       osd_nodes = osd_nodes.take(2)
     end
 
-    mon_nodes = select_nodes_for_role(nodes, "ceph-mon", "controller")
+    mon_nodes = select_nodes_for_role(nodes, "ceph-mon", "storage")
 
     if mon_nodes.size < 3
-      mon_nodes_storage = select_nodes_for_role(nodes, "ceph-mon", "storage")
-      mon_nodes_all = select_nodes_for_role(nodes, "ceph-mon")
-      mon_nodes = [mon_nodes, mon_nodes_storage, mon_nodes_all].flatten.uniq{ |n| n.name }
+      mon_nodes_more = select_nodes_for_role(nodes, "ceph-mon").reject do |n|
+        n.intended_role == "controller"
+      end
+      mon_nodes = [mon_nodes, mon_nodes_more].flatten.uniq(&:name)
     end
     mon_nodes = mon_nodes.take(mon_nodes.length > 2 ? 3 : 1)
 
-    mds_node = select_nodes_for_role(nodes, "ceph-mds", "controller").first
+    mds_node = select_nodes_for_role(nodes, "ceph-mds").reject do |n|
+      n.intended_role == "controller" or osd_nodes.include? n
+    end.first
+    if mds_node.nil?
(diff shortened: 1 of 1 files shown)


crowbar_framework/app/models/ceph_service.rb
```diff
@@ -122,13 +122,13 @@ def create_proposal
...
       osd_nodes_all = select_nodes_for_role(nodes, "ceph-osd")
-      # avoid controllers if possible (will likely be used for mon)
+      # avoid controllers if possible (ceph should not be used with openstack roles)
       osd_nodes_no_controller = osd_nodes_all - osd_nodes_controller
       osd_nodes = [osd_nodes, osd_nodes_no_controller, osd_nodes_all].flatten.uniq{ |n| n.name }
       osd_nodes = osd_nodes.take(2)
     end
 
-    mon_nodes = select_nodes_for_role(nodes, "ceph-mon", "controller")
+    mon_nodes = select_nodes_for_role(nodes, "ceph-mon", "ceph")
```
cmurphy: Is "ceph" a valid node role? It's not selectable in the UI. Do you mean "storage"?

crowbar_framework/app/models/ceph_service.rb
```diff
@@ -137,9 +137,9 @@ def create_proposal
     end
     mon_nodes = mon_nodes.take(mon_nodes.length > 2 ? 3 : 1)
 
-    mds_node = select_nodes_for_role(nodes, "ceph-mds", "controller").first
+    mds_node = select_nodes_for_role(nodes, "ceph-mds", "storage").first
 
-    radosgw_node = select_nodes_for_role(nodes, "ceph-radosgw", "controller").first
+    radosgw_node = select_nodes_for_role(nodes, "ceph-radosgw", "storage").first
```
cmurphy: I think radosgw is supposed to go on the controller

cmurphy (changes requested):

crowbar_framework/app/models/ceph_service.rb
jsuchome: Yes, I meant storage, you are right

crowbar_framework/app/models/ceph_service.rb
jsuchome: I think this https://www.suse.com/documentation/ses-4/book_storage_admin/data/sysreq_mon.html talks about all ceph roles. But maybe @tserong can correct us.

cmurphy: As an API service I would expect it to be fine running on the controller, and I would not expect it to cohabitate a host with the storage services. The docs here https://www.suse.com/documentation/ses-4/book_storage_admin/data/ceph_install_crowbar_deploy_nodes.html recommend installing it on a dedicated node, so I guess as long as this is separate from the ceph-mon and ceph-osd nodes it will be fine?

jsuchome: Yes, I agree it should not actually hurt. But let's wait what Tim has to say

cmurphy: :+1:

tserong (approved):

crowbar_framework/app/models/ceph_service.rb
tserong: The main thing the requirements are trying to say is to not mix the workload with the ceph services. Additionally, ideally, all ceph services (be they MONs, OSDs or RGWs) would be on separate hosts.

@cmurphy as you say, RGW is probably OK running on a controller; I imagine that's not going to be heavily loaded anyway. But, IMO all ceph roles should default to landing on nodes that were declared for storage use. The user can always shunt them elsewhere should they so chose.

jsuchome: Let's also have in mind, this is just the automatic pre-deployment; user is still allowed to change it

Itxaka:
So this breaks the normal CI run but probably fixes the HA ci run \~_\~

jsuchome:
@Itxaka fixed (or rather "fixed")

Itxaka (approved):
Looks good with the last change, thanks!

cmurphy (approved):

jsuchome:
Adding do-not-merge, as the deployment still does not work for HA setup (because there are not enough 'storage' intended roles)

jsuchome:
[Fix automatic deployment of ceph roles](https://trello.com/c/F21pgJnm/310-fix-automatic-deployment-of-ceph-roles)

cmurphy:
Still :+1: from me assuming CI passes

Itxaka (approved):

crowbar_framework/app/models/ceph_service.rb
dirkmueller: @jsuchome I believe we documented somewhere that mixing openstack roles with ceph roles is unsupported, so avoiding that via constraints checks should be good for now (it might break existing customers violating that though).

PR #95, opened 2017-06-07
Do not allow ceph-mon and ceph-osd deployment to cluster members.

jsuchome:
Previous check for 'controller' intended role is not enough: non-controller
nodes might be added to the cluster automatically by mkcloud.

Followup after https://github.com/crowbar/crowbar-ceph/pull/94


Itxaka (approved):
This should definitively work, thanks!

cmurphy (approved):
This seems to fix it for me

jsuchome:
Hm, the HA job fails the usual way. @cmurphy @Itxaka did you test it with some real setups?

cmurphy:
I thought I did :(

cmurphy:
Re-upping my approval :+1:

s-t-e-v-e-n-k (approved):

PR #96, opened 2017-06-07
Put ceph-mds on the founder if necessary

cmurphy:


jsuchome:
It might be enough to do the same as for mon/osd:
```
      mds_node = select_nodes_for_role(nodes, "ceph-mds").reject do |n|
￼        n.intended_role == "controller" ||  osd_nodes.include? n || n.roles.include?("pacemaker-cluster-member")
￼      end.first
```

For smaller deployments, we can end without any ceph-mds, though

crowbar_framework/app/models/ceph_service.rb
```diff
@@ -142,7 +143,10 @@ def create_proposal
       n.intended_role == "controller" or osd_nodes.include? n
     end.first
     if mds_node.nil?
-      mds_node = select_nodes_for_role(nodes, "ceph-mds", "controller").first
+      mds_node = select_nodes_for_role(nodes, "ceph-mds", "controller")
+      mds_node = mds_node.select do |n|
+        CrowbarPacemakerHelper.is_cluster_founder?(n)
```
cmurphy, line 148: This actually not going to work because in our CI we only have one node intended as a controller (which is usually the founder too) and so this case will never be hit.

cmurphy:
Abandoning in favor of #95

PR #97, opened 2017-06-13
Make sure ceph-mds and ceph-calamari are not deployed to the cluster node.

jsuchome:
We cannot use standard `select_nodes_for_role` because it
limits the search for available nodes with 'count' number.

### DIFF
--- a/crowbar_framework/app/models/ceph_service.rb
+++ b/crowbar_framework/app/models/ceph_service.rb
@@ -106,6 +106,27 @@ class CephService < PacemakerServiceObject
     answer
   end
 
+  # Helper to find a suitable node for ceph-mds role.
+  # ceph-mds conflicts with ceph-osd, and should not be assigned to the cluster node
+  def select_node_for_mds_role(nodes)
+    # do not modify array given by caller
+    mds_nodes = nodes.dup
+    mds_nodes.delete_if(&:nil?)
+    mds_nodes.reject! do |n|
+      node_is_valid_for_role(n.name, "ceph-mds")
+    end
+
+    mds_node = mds_nodes.find { |n| n.intended_role != "controller" }
+    if mds_node.nil? && !mds_nodes.empty?
+      mds_node = mds_nodes.first
+      @logger.debug("Not enough nodes: putting ceph-mds on controller node (unsupported scenario)")
+    end
+    if mds_node.nil?
+      @logger.warn("Not enough nodes: there seems to be no node for ceph-mds role!")
+    end
+    mds_node
+  end
+
   def create_proposal
     @logger.debug("Ceph create_proposal: entering")
     base = super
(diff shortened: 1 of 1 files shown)


jsuchome:
Alternatively, we can just remove that `if mds_node.nil?` block completely, it's already unlikely that it find some node that is intended controller, but not in cluster...

jsuchome:
Oh. Manila needs ceph-mds to be deployed... (see HA job failure)

tserong (dismissed):

sjamgade (dismissed):

jsuchome:
[Fix automatic deployment of ceph roles](https://trello.com/c/F21pgJnm/310-fix-automatic-deployment-of-ceph-roles)

crowbar_framework/app/models/ceph_service.rb
```diff
@@ -106,6 +106,27 @@ def proposal_dependencies(role)
...
   end
 
+  # Helper to find a suitable node for ceph-mds role.
+  # ceph-mds conflicts with ceph-osd, and should not be assigned to the cluster node
+  def select_node_for_mds_role(nodes)
+    # do not modify array given by caller
+    mds_nodes = nodes.dup
+    mds_nodes.delete_if(&:nil?)
+    mds_nodes.reject! do |n|
+      node_is_valid_for_role(n.name, "ceph_mds") || n.roles.include?("pacemaker-cluster-member")
```
jsuchome: This is very confusing. I assume the correct name for `node_is_valid_for_role` should be `node_is_not_valid_for_role` ... it's defined here https://github.com/crowbar/crowbar-core/blob/master/crowbar_framework/app/models/service_object.rb#L442

tserong (dismissed):
I'd still like one other pair of eyes that aren't mine on the `node_is_valid_for_role(n.name, "ceph-mds")` piece :-) but LGTM.

crowbar_framework/app/models/ceph_service.rb
```diff
@@ -166,9 +166,9 @@ def create_proposal
 
     # Any spare node after allocating mons and osds is fair game
     # to automatically use as the calamari server
-    calamari_nodes = select_nodes_for_role(nodes, "ceph-calamari")
+    calamari_nodes = select_nodes_for_role(nodes - osd_nodes - mon_nodes, "ceph-calamari")
     calamari_nodes.reject! do |n|
```
rhafer, line 173: Doesn't this suffer from exactly the same problem as the mds node selection? I.e. you might endup with no remaining nodes if the node returned by ```select_nodes_for_role``` is a pacemaker member? Or am I missing something?

rhafer: BTW, wouldn't it be simpler to just reject all cluster members before calling "select_nodes_for_role" and "select_node_for_mds_role" ?

rhafer (changes requested):
I think for the 1st commit you should elaborate a bit more on why ```select_nodes_for_role``` is problematic. It took me a while to actually figure out which "count" number the commit message is referring to.

crowbar_framework/app/models/ceph_service.rb
jsuchome: > Doesn't this suffer from exactly the same problem as the mds node selection

You are right (although with calamari it is not such problem when it is not deployed at all)

> reject all cluster members before calling 

That's a good point

rhafer (approved):
I think this is ok now

PR #98, opened 2017-06-19
Only deploy ceph roles to SP2 nodes.

jsuchome:
We do not provide any SES product for SLES12SP3 nodes.

See also https://github.com/crowbar/crowbar-core/pull/1249

Requires https://github.com/crowbar/crowbar-openstack/pull/1012 so that cinder-volume is not deployed to ceph node

### DIFF
--- a/crowbar_framework/app/models/ceph_service.rb
+++ b/crowbar_framework/app/models/ceph_service.rb
@@ -49,7 +49,7 @@ class CephService < PacemakerServiceObject
           "unique" => false,
           "count" => 1,
           "platform" => {
-            "suse" => "/^12.*/",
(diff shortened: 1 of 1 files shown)


aplanas (approved):
Do we need to revert this after we have a new version of SES?

jsuchome:
@aplanas if there'd be a version of SES that does support crowbar, sure. But I kind of doubt this is gonna happen: 
http://docserv.suse.de/documents/Storage_5/ses-deployment/single-html/#sec.depl.docupdates.5

rsalevsky (approved):

tserong (approved):

PR #100, opened 2017-06-30
[stable/4.0] Make sure ceph-mds is deployed (needed for manila).

jsuchome:
As we need only one ceph-mds node, let's take it first and we can
both make sure that there's one deployed and skip those complicated
checks for conflicts using the extra method.

(cherry picked from commit 7f61904ed66837265d210a483eb57031560b5e85)

### DIFF
--- a/crowbar_framework/app/models/ceph_service.rb
+++ b/crowbar_framework/app/models/ceph_service.rb
@@ -106,27 +106,6 @@ class CephService < PacemakerServiceObject
     answer
   end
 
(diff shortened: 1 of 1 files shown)


aplanas:
We need that for gating SOC7

skazi0 (approved):
lgtm

aplanas (approved):

s-t-e-v-e-n-k (approved):

toabctl (approved):

tserong (approved):

PR #102, opened 2017-08-02
[4.0] Decrease the amount of node saves

scottwulf:
Rebase of PR #80
Only save the node when attribute(s) have changed
Do not fetch monitor secret if we're the master

### DIFF
--- a/chef/cookbooks/ceph/recipes/mon.rb
+++ b/chef/cookbooks/ceph/recipes/mon.rb
@@ -52,6 +52,8 @@ unless File.exist?("/var/lib/ceph/mon/ceph-#{mon_name}/done")
       add_key.run_command
       add_key.error!
 
+      # no need to check if the attribute is already set: it's part of the
+      # only_if
       node.set["ceph"]["monitor-secret"] = monitor_key
       node.save
     end
@@ -85,10 +87,11 @@ unless File.exist?("/var/lib/ceph/mon/ceph-#{mon_name}/done")
       add_key.run_command
       add_key.error!
 
+      # no need to check if the attribute is already set: it's part of the
       node.set["ceph"]["monitor-secret"] = monitor_key
       node.save
     end
-    only_if { node["ceph"]["monitor-secret"].empty? }
+    only_if { node["ceph"]["monitor-secret"].empty? && !node[:ceph][:master] }
     notifies :run, "execute[ceph-mon mkfs]", :immediately
   end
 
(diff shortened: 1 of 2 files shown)


nicolasbock (dismissed):

chef/cookbooks/ceph/recipes/mon.rb
```diff
@@ -85,10 +87,11 @@
       add_key.run_command
       add_key.error!
 
+      # no need to check if the attribute is already set: it's part of the
```
Itxaka: comment not finished

chef/cookbooks/ceph/recipes/osd.rb
```diff
@@ -127,7 +137,7 @@
         end
         device["device"] = d.name
         node.set["ceph"]["osd_devices"].push(device)
-        node.save
+        dirty = true
```
Itxaka, line 140: We were saving here explicitly for a reason, probably becasue we are using this values `["ceph"]["osd_devices"]` immediately below.

Are we sure this is still working as before? I would say yes, as its only used in this recipe, but would like confirmation that this has been tested properly and the devices are being picked up in the recipe afterwards without saving the node.

Itxaka (dismissed):
Looks good, just a question and a nit

Itxaka (commented):
Looks good, just a question and a nit

chef/cookbooks/ceph/recipes/osd.rb
```diff
@@ -71,7 +71,7 @@
   if is_crowbar?
     dirty = false
 
-    node.set["ceph"]["osd_devices"] ||= []
+    node.set["ceph"]["osd_devices"] = [] if node["ceph"]["osd_devices"].nil?
```
nicolasbock: Same question as is in https://github.com/crowbar/crowbar-ceph/pull/103. Why the change?

chef/cookbooks/ceph/recipes/osd.rb
scottwulf: Yes, the array of devices is enumerable (i.e. usable) immediately below even after moving node.save to the end of the file. The node.save is only needed to persist the values for subsequent executions of the cookbook.

Itxaka: Thanks!

Itxaka (approved):

nicolasbock (approved):

PR #103, opened 2017-08-04
Decrease the amount of node saves

scottwulf:
Fix multiline comment
Fix array initialization (was mistakenly initializing as a hash)

### DIFF
--- a/chef/cookbooks/ceph/recipes/osd.rb
+++ b/chef/cookbooks/ceph/recipes/osd.rb
@@ -71,7 +71,7 @@ else
   if is_crowbar?
     dirty = false
 
-    node.set["ceph"]["osd_devices"] ||= []
+    node.set["ceph"]["osd_devices"] = [] if node["ceph"]["osd_devices"].nil?
     min_size_blocks = node["ceph"]["osd"]["min_size_gb"] * 1024 * 1024 * 2
     unclaimed_disks = BarclampLibrary::Barclamp::Inventory::Disk.unclaimed(node).sort.select { |d| d.size >= min_size_blocks }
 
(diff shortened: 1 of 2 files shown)


chef/cookbooks/ceph/recipes/osd.rb
```diff
@@ -71,7 +71,7 @@
   if is_crowbar?
     dirty = false
 
-    node.set["ceph"]["osd_devices"] ||= []
+    node.set["ceph"]["osd_devices"] = [] if node["ceph"]["osd_devices"].nil?
```
nicolasbock, line 74: Sorry, I am still not completely comfortable with the `||=` notation, but do I interpret your change correctly that it doesn't change anything in terms of what the code does? If that's the case, then why change it?

scottwulf: This it to revert the change I accepted from the original PR. I too was tricked by the ||= and initial tests did not reveal an issue. Ceph is currently broken.

||= does nothing currently because node.set[] is true therefore = is never executed, leaving the variable unset [and as a hash], and after all of the pushes it's value is a nested hash, contains only the last device, and looks like {"push": { "name": "last_device_pushed" }}.

I chose to leave the number of changes to a minimum and simply revert the change. There is another option. If we as an organization are not adverse to changing code that already works for the sake of making it written more properly, in osd.rb, I will change "node.set[] ||= []" to "node[] ||= []" and "node.set[].push()" to "node[].push()" and create a new PR.

Both options have been tested, please let me know which path to proceed. Accept this PR or let me know to create a new PR. I am fine with either approach.

Itxaka: wow, great finding @scottwulf !

Im ok with this change, seems better than the push.

Itxaka (approved):

chef/cookbooks/ceph/recipes/osd.rb
nicolasbock: Unbelievable! Thanks for the detailed explanation @scottwulf! I like this change!

nicolasbock (approved):

PR #107, opened 2018-02-07
get_osd_id: silence errors

bmwiedemann:
https://bugzilla.suse.com/show_bug.cgi?id=1073237#c4 & 5

When creating OSDs, it waits for for get_osd_id to return a value
which caused ugly error messages in the logs:
cat: /var/lib/ceph/osd/ceph-7/var/lib/ceph/tmp/mnt.lAtztf/whoami: No such file or directory

### DIFF
--- a/chef/cookbooks/ceph/libraries/default.rb
+++ b/chef/cookbooks/ceph/libraries/default.rb
@@ -164,7 +164,7 @@ end
 
 def get_osd_id(device)
   osd_path = %x[mount | grep #{device} | awk '{print $3}'].tr("\n","")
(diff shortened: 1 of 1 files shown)


jsuchome:
Shouldn't we add a check for that file presence?

bmwiedemann:
what would be the advantage? cat already checks for presence...
or do you mean a separate checking function?

jsuchome:
Just thought it might be nicer and more readable, but probably is not really needed

nicolasbock (commented):
It's definitely compact @bmwiedemann :wink:. I still don't quite understand how `stderr` ends up in the log file.

JanZerebecki (approved):

PR #108, opened 2018-03-06
Do not write fstab entries (bsc#1073237)

bmwiedemann:
because it can have races with ceph-disk activate moving mounts
https://bugzilla.suse.com/show_bug.cgi?id=1073237

Also it made the whole OSD allocation less dynamic than it could be.

backport of #105

### DIFF
--- a/chef/cookbooks/ceph/recipes/osd.rb
+++ b/chef/cookbooks/ceph/recipes/osd.rb
@@ -218,11 +218,6 @@ else
           dirty = true
         end
 
-        execute "Writing Ceph OSD device mappings to fstab" do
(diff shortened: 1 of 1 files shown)


nicolasbock (changes requested):
Please `cherry-pick -x` so we get the commit hash as references.

bmwiedemann:
@nicolasbock done

dirkmueller (approved):

nicolasbock (approved):

dirkmueller:
hmm, the segfault in ceph doesn't look good :/

tserong:
> hmm, the segfault in ceph doesn't look good :/

I'm looking at that. It's almost certainly unrelated to this change.

PR #109, opened 2018-03-07
Hide ceph barclamp from default view (bsc#1076748)

dirkmueller:
We need to see how we can do an external SES integration, but
for now we need to remove the visibility of the internal SES
support as it isn't working anymore for Cloud 8.

### DIFF
--- a/ceph.yml
+++ b/ceph.yml
@@ -20,7 +20,7 @@ barclamp:
   display: 'Ceph'
   description: 'Distributed object store and file system'
   version: 0
-  user_managed: true
+  user_managed: false
   member:
     - 'openstack'
     - 'suse_enterprise_storage'


rsalevsky:
What about just adding a UI warning that only external ceph is supported? I know it's not a good solution but we also some other cases where we have to adjust the barclamp to reflect the current support status.

toabctl (approved):
I'm fine with this solution.

vuntz:
@rsalevsky @dirkmueller any reason to not simply drop the crowbar-ceph package from the media?

toabctl:
> any reason to not simply drop the crowbar-ceph package from the media?

@vuntz I guess that would require multiple changes in different barclamps, no?. Hiding it sounds like the easiest solution to get this into GMC.

vuntz (approved):
> @vuntz I guess that would require multiple changes in different barclamps, no?

Hrm, we do include `ceph::keyring` in three locations, but this should only be happening for deployments done by the ceph barclamp.

That being said, we're so close to GMC that I agree that it's risky and hiding is easier...

rsalevsky (approved):

